PR #53, opened 2016-03-22
[DOC] add core concepts docs

tessalt:
now that I have written these, I don't know that they're really that useful. I also have no idea what to say about collections. 

@minasmart @richgilbank @danielpatricio @harismahmood89

### DIFF
--- a/core-concepts/index.md
+++ b/core-concepts/index.md
@@ -1,0 +1,23 @@
+---
+layout: default
+---
+# Core Concepts
+
+The Shopify JavaScript Buy SDK uses the Shopify API to connect to your shop, and is based
+on the same core concepts.
+
+## Products and Variants
+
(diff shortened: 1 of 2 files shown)


minasmart:
I was thinking that a doc on `models` may be a more general place to document. You're right about a lot of this being self-explanatory, but what isn't, is that models are mostly ES5 getters and setters.
It's not obvious anywhere how model attributes work, vs how to see true internal state that the API delivers. I may start working on that soon if you're tied up :+1:

harismahmood89:
@tessalt i think this is super valuable - it may all be obvious to us but I can totally see it being helpful to get a quick overview of what to look out for when working with the SDK. :+1:

tessalt:
closing this for now, we can revisit later.

Issue #34, opened 2016-03-18
unhelpful logging on deploy


tessalt:
![image](https://cloud.githubusercontent.com/assets/1789029/13893251/3ac7d732-ed33-11e5-98fd-b27caa1b6b81.png)

etc

PR #54, opened 2016-03-22
disable progress bar in shipit

tessalt:
I think this is what is causing the useless random logging in shipit as seen in https://github.com/Shopify/js-buy-sdk/issues/34

@minasmart @richgilbank


minasmart:
LGTM

Issue #51, opened 2016-03-22
v0.2.0 punch list


minasmart:
- [x] As a quick stopgap, provide a convenience method to fetch the most recent cart if it exists, or create one. `client.fetchRecentCart() -> Promise | CartModel`
- [x] Allow full domains, over just myshopify domains. Going forward we will be supporting different TLDs.
- [x] `cart.lineItems` should be list of class instances, rather than a regular hash representation. Underscores should go away.
- [x] Add `cart.lineItemCount` for total number of items in cart, not just `lineItems.length` (Inspired by #57 and https://github.com/Shopify/js-buy-sdk/pull/57/files#diff-2252a2513e98df574dd8383c6bbfc5dbR242)
- [x] Images sizes (see #10)
- [x] Safely handle no product images. Currently, things fail out on IE9 when there are no images.
- [x] Add Version Header to requests
- [x] Add version mark to artifact.
- [ ] ~~Support line item properties~~

@alexaitken, @richgilbank, @tessalt, @harismahmood89: If you know of things that should be changed going forward please add them here.

minasmart:
We did all this a while ago! Woo!

PR #57, opened 2016-03-28
update cart example

harismahmood89:
Updates cart example to add the following and to match SDK documentation:
- persist cart
- cart tab update: increment, show, hide
- cart hide when items count = 0
- update functions to use getters

please review: @minasmart

### DIFF
--- a/examples/cart/index.html
+++ b/examples/cart/index.html
@@ -61,6 +61,14 @@
   <!-- .cart end -->
 
 
+  <!-- .cart-tab start -->
+  <button class="btn btn--cart-tab">
+    <span class="btn__counter"></span>
+    <svg xmlns="http://www.w3.org/2000/svg" class="icon-cart icon-cart--side" viewBox="0 0 25 25" enable-background="new 0 0 25 25"><g fill="#fff"><path d="M24.6 3.6c-.3-.4-.8-.6-1.3-.6h-18.4l-.1-.5c-.3-1.5-1.7-1.5-2.5-1.5h-1.3c-.6 0-1 .4-1 1s.4 1 1 1h1.8l3 13.6c.2 1.2 1.3 2.4 2.5 2.4h12.7c.6 0 1-.4 1-1s-.4-1-1-1h-12.7c-.2 0-.5-.4-.6-.8l-.2-1.2h12.6c1.3 0 2.3-1.4 2.5-2.4l2.4-7.4v-.2c.1-.5-.1-1-.4-1.4zm-4 8.5v.2c-.1.3-.4.8-.5.8h-13l-1.8-8.1h17.6l-2.3 7.1z"></path><circle cx="9" cy="22" r="2"></circle><circle cx="19" cy="22" r="2"></circle></g></svg>
+  </button>
+  <!-- .cart-tab end -->
+
+
   <script id="cart-item-template" type="text/template">
     <div class="cart-item">
       <div class="cart-item__img"></div>
--- a/examples/cart/index.js
+++ b/examples/cart/index.js
@@ -9,11 +9,23 @@ $(function() {
   });
 
   var cart;
-  client.createCart().then(function (newCart) {
-    cart = newCart;
-  });
+  var cartLineItemCount;
+  if(localStorage.getItem('lastCartId')) {
+    client.fetchCart(localStorage.getItem('lastCartId')).then(function(remoteCart) {
+      cart = remoteCart;
+      cartLineItemCount = cart.lineItems.length;
+      renderCartItems();
+    });
+  } else {
+    client.createCart().then(function (newCart) {
+      cart = newCart;
+      localStorage.setItem('lastCartId', cart.id);
+      cartLineItemCount = 0;
+    });
+  }
 
-  var lineItemImages = {};
+
+
 
   /* Fetch product and init
   ============================================================ */
@@ -35,6 +47,12 @@ $(function() {
 
     attachQuantityIncrementListeners(product);
     attachQuantityDecrementListeners(product);
+
+    updateCartTabButton();
+
+    attachCheckoutButtonListeners();
+
+    closeCart();
 
   });
 
@@ -99,10 +117,8 @@ $(function() {
   function attachBuyButtonListeners(product) {
     $('.buy-button').on('click', function (event) {
       event.preventDefault();
-      var image = product.selectedVariantImage;
       var id = product.selectedVariant.id;
-      lineItemImages[id] = image.src;
-      addVariantToCart(product, product.selectedVariant, 1);
+      addVariantToCart(product.selectedVariant, 1);
     });
   }
 
@@ -110,7 +126,7 @@ $(function() {
   ============================================================ */
   function attachQuantityIncrementListeners(product) {
     $('.cart').on('click', '.quantity-increment', function() {
-      var variantId = parseInt($(this).attr('data-variant-id'));
+      var variantId = parseInt($(this).attr('data-variant-id'), 10);
       var variant = product.variants.filter(function (variant) {
         return (variant.id === variantId);
       })[0];
@@ -118,7 +134,7 @@ $(function() {
       $(this).closest('.cart-item').addClass('js-working');
       $(this).attr('disabled', 'disabled');
 
-      addVariantToCart(product, variant, 1);
+      addVariantToCart(variant, 1);
     });
   }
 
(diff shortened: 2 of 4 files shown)


examples/cart/index.html
```diff
@@ -61,6 +61,14 @@ <h2 class="cart-title">Your cart</h2>
   <!-- .cart end -->
 
 
+  <!-- .cart-tab start -->
+  <button class="btn btn--cart-tab">
+    <div class="btn__counter"></div>
+    <svg xmlns="http://www.w3.org/2000/svg" class="icon-cart icon-cart--side" viewBox="0 0 25 25" enable-background="new 0 0 25 25"><g fill="#fff"><path d="M24.6 3.6c-.3-.4-.8-.6-1.3-.6h-18.4l-.1-.5c-.3-1.5-1.7-1.5-2.5-1.5h-1.3c-.6 0-1 .4-1 1s.4 1 1 1h1.8l3 13.6c.2 1.2 1.3 2.4 2.5 2.4h12.7c.6 0 1-.4 1-1s-.4-1-1-1h-12.7c-.2 0-.5-.4-.6-.8l-.2-1.2h12.6c1.3 0 2.3-1.4 2.5-2.4l2.4-7.4v-.2c.1-.5-.1-1-.4-1.4zm-4 8.5v.2c-.1.3-.4.8-.5.8h-13l-1.8-8.1h17.6l-2.3 7.1z"></path><circle cx="9" cy="22" r="2"></circle><circle cx="19" cy="22" r="2"></circle></g></svg>
```
minasmart: Should this live in its own file maybe? Keep large blobs out of the index?

I don't feel strongly either way.

examples/cart/index.js
```diff
@@ -9,9 +9,17 @@ $(function() {
   });
 
   var cart;
-  client.createCart().then(function (newCart) {
-    cart = newCart;
-  });
+  if(localStorage.lastCartId) {
```
minasmart: use `getItem` and `setItem`. If we add tests to this, those can be stubbed. Testing that properties are accessed is harder.

examples/cart/index.html
harismahmood89: It's such a small example that I think we should keep it all in the same place. Keeps the example as straight forward as possible

minasmart: :+1:

examples/cart/index.js
```diff
@@ -99,10 +113,10 @@ $(function() {
   function attachBuyButtonListeners(product) {
     $('.buy-button').on('click', function (event) {
       event.preventDefault();
-      var image = product.selectedVariantImage;
       var id = product.selectedVariant.id;
-      lineItemImages[id] = image.src;
-      addVariantToCart(product, product.selectedVariant, 1);
+      var image = product.selectedVariant.image.src;
+      lineItemImages[id] = image;
```
minasmart: You're not actually consuming these values anywhere any more, so :fire:

examples/cart/index.js
```diff
@@ -110,15 +124,15 @@ $(function() {
   ============================================================ */
   function attachQuantityIncrementListeners(product) {
     $('.cart').on('click', '.quantity-increment', function() {
-      var variantId = parseInt($(this).attr('data-variant-id'));
-      var variant = product.variants.filter(function (variant) {
+      var variantId = parseInt($(this).attr('data-variant-id'), 10);
+      var variant = product.attrs.variants.filter(function (variant) {
```
minasmart: use `product.variants`

examples/cart/index.js
```diff
@@ -126,15 +140,15 @@ $(function() {
   ============================================================ */
   function attachQuantityDecrementListeners(product) {
     $('.cart').on('click', '.quantity-decrement', function() {
-      var variantId = parseInt($(this).attr('data-variant-id'));
-      var variant = product.variants.filter(function (variant) {
+      var variantId = parseInt($(this).attr('data-variant-id'), 10);
+      var variant = product.attrs.variants.filter(function (variant) {
```
minasmart: `product.variants`

examples/cart/index.js
```diff
@@ -151,48 +165,59 @@ $(function() {
...
-      $cartItemContainer.append($cartLineItems);
-
-      setTimeout(function () {
-        $cartItemContainer.find('.js-hidden').removeClass('js-hidden');
-      }, 0)
-
-      $('.cart .pricing').text(formatAsMoney(cart.subtotal));
+    var itemExists;
+    if (cart.lineItems.length > 0) {
+      itemExists = cart.lineItems.filter(function (cartItem) {
```
minasmart: Instead of doing this filter, maybe check the length of `lineItems` and see if length grows. I think that's what's trigger your animation anyway, right? Appending to the list?

examples/cart/index.js
```diff
@@ -201,4 +226,32 @@ $(function() {
...
+    });
+  }
+
+  /* Update cart tab button
+  ============================================================ */
+  function updateCartTabButton() {
+    if (cart.lineItems.length > 0) {
+      var totalItems = cart.lineItems.reduce(function(total, item) {
+        return total + item.quantity;
+      }, 0);
```
minasmart: Made a note to move this work to the cart: https://github.com/Shopify/js-buy-sdk/issues/51

minasmart:
Couple small issues, then set sail! :boat:

richgilbank:
Let's expand these out a bit now that the context is all available from being open source. @Shopify/channels-fed can one of you take a look at this? 
Background: this is the example for a cart, using the JS Buy SDK - it's styled to be similar to the Buy Button's embedded cart, but work sans-iframe. Browser support is aiming for IE9+, though CORS support is temporarily limited to IE10+.

minasmart:
IE9 CORS currently predicated on: https://github.com/Shopify/shopify/pull/68257

examples/cart/index.html
```diff
@@ -61,6 +61,14 @@ <h2 class="cart-title">Your cart</h2>
   <!-- .cart end -->
 
 
+  <!-- .cart-tab start -->
+  <button class="btn btn--cart-tab">
+    <div class="btn__counter"></div>
```
nwtn: Technically only phrasing content can live inside a `button`, which means this would need to be a `span` instead of a `div`. I’ve never seen it cause an actual problem, but if switching to `span` with `display: block` doesn’t cause issues it may be safer.

examples/cart/index.js
```diff
@@ -151,48 +165,59 @@ $(function() {
...
+    if (cart.lineItems.length > 0) {
+      itemExists = cart.lineItems.filter(function (cartItem) {
+        return (cartItem.variant_id === variant.id);
+      })[0];
+    }
 
+    cart.addVariants({ variant: variant, quantity: quantity }).then(function() {
+      renderCartItems(itemExists);
     }).catch(function (errors) {
       console.log("Fail");
```
nwtn: :cop: i think we normally do single quotes

harismahmood89:
@minasmart ready to go. @nwtn thanks for the review, homie

examples/cart/index.js
```diff
@@ -9,11 +9,23 @@ $(function() {
   });
 
   var cart;
-  client.createCart().then(function (newCart) {
-    cart = newCart;
-  });
+  var cartLineItemCount;
+  if(localStorage.lastCartId) {
```
minasmart: always use `getItem` and `setItem`

minasmart:
:+1: just that one thing about using `localStorage` getters and setters. I'd like to have tests for this at some point and getters and setters let us stub native behaviours.

harismahmood89:
Updated all `localStorage` stuff to use getters and setters.

examples/cart/index.js
```diff
@@ -151,48 +167,59 @@ $(function() {
...
+  function renderCartItems() {
+    var $cartItemContainer = $('.cart-item-container');
+    var totalPrice = 0;
+
+    $cartItemContainer.empty();
+    var lineItemEmptyTemplate = $('#cart-item-template').html();
+    var $cartLineItems = cart.lineItems.map(function (lineItem, index) {
+      var $lineItemTemplate = $(lineItemEmptyTemplate);
+      var itemImage = lineItem.image.src;
+      $lineItemTemplate.find('.cart-item__img').css('background-image', 'url(' + itemImage + ')');
```
tylerball: are you avoiding using a templating system here for any reason?

harismahmood89: Just trying to keep the example as simple and small as possible. The focus of the example is meant to be the interactions with the SDK, so I didn't want to add in anything fancy with templating. I'm open to hearing other options though - whatever makes the example as simple and digestible as possible. 

@richgilbank thoughts?

harismahmood89: just chatted with tyler in person - verdict was that it makes sense to not add an additional dependency for an example.

PR #58, opened 2016-03-29
[FIX] Support the network layer in IE9

minasmart:
Depends on: https://github.com/Shopify/shopify/pull/68257

Uses XDRs with *special* authorization for *special* IE9, since CORS is not supported through XHR in that environment.

**NOTE:** IE9 has very limited error handling and debugging capability with this approach. This constitutes a bare minimum of functionality.

**OTHER NOTE:** This does not fix mixed schemes. At this point, IE9 will have to be supported on `https` scheme websites only. Fixing this caveat requires work in Shopify Core, and does not constitute a release.


minasmart:
CC: @richgilbank @tessalt

tessalt:
very cool. And by cool I mean IE9 is terrible, but gj.

PR #70, opened 2016-04-04
Adding the readme for the examples section

richgilbank:
Adds a readme to the cart example directory since it wasn't documented anywhere previously: https://ecommerce.shopify.com/c/shopify-apis-and-technology/t/cart-example-not-working-334493

I think this is the right place to put this, but open to suggestions @tessalt @harismahmood89

### DIFF
--- a/examples/cart/README.md
+++ b/examples/cart/README.md
@@ -1,0 +1,10 @@
+### JS Buy SDK Examples
+
+To run this example locally, you must first install the SDK's
+dependencies and start a server to serve these files from:
+
+```bash
+npm install && npm run examples
+```
+
+Once it's running, open your browser to `http://localhost:4200/examples/cart/` to view the example.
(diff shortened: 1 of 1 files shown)


richgilbank:
Bump 😄

tessalt:
does `ENV=examples` still need to be set?

richgilbank:
Not that I know of, but I've never seen it needed previously, so you know more about it than I!

harismahmood89:
Do we want to add more details about what the requirements are? The person in the ticket example is quite confused

richgilbank:
Any requirements in particular you have in mind?

tessalt:
@richgilbank if you just pull down the repo and follow this instruction does it work?

richgilbank:
It does, but I figured we should add a note as to what to open the browser to, so I've done that. GH isn't picking up new commits at the moment though, will try pushing the change tomorrow.

minasmart:
npm run examples sets the end variable. This looks good :smile: :+1:

harismahmood89:
@richgilbank maybe just listing off node and browser requirements?

minasmart:
LGTM.

richgilbank:
Good call @harismahmood89 - added.

Issue #69, opened 2016-04-04
Unable to initialise cart with variants. Documentation example not working.


dmitrybelyakov:
Hi, I'm following official documentation from [Javascript Buy SDK Guide](http://shopify.github.io/js-buy-sdk/guide/)  - initialising cart with variants in fact does nothing for me, apart from creating an empty cart:

```javascript
//these are actual products
var productId = 5863604999;
var variantId = 18525309447;

//code from the docs: http://shopify.github.io/js-buy-sdk/guide/
shopClient.createCart({id: variantId, quantity: 1}).then(function (cart) {
    console.log('CREATED CART:', cart.lineItems.length); //0
});


//doing it like this doesn't help either
shopClient.fetchProduct(productId).then(function (product) {
    shopClient.createCart({id: product.selectedVariant.id, quantity: 1}).then(function (cart) {
        console.log('CREATED CART:', cart.lineItems.length); //0
    });
 });
```

using: [shopify-buy.polyfilled.globals.min.js](http://sdks.shopifycdn.com/js-buy-sdk/latest/shopify-buy.polyfilled.globals.min.js)

dmitrybelyakov:
The error I'm getting here is this:

```
Uncaught (in promise) TypeError: Cannot read property 'image' of undefined
```

Which seems to happen [at this line](https://github.com/Shopify/js-buy-sdk/blob/master/src/models/cart-model.js#L103) in cart model.

richgilbank:
Hey @dmitrybelyakov, thanks for letting us know. Looks like it's expected to be under the `line_items` key in the attributes passed to `createCart`. This should work:
```javascript
shopClient.createCart({ line_items: [{ id: variantId, quantity: 1 }] }).then(function(cart) { ...
```
Worth noting though, that this won't automatically fetch the variant image or other details, in case you need those to display a cart.
I'll update the docs - please let us know if you come across anything else. Thanks!

dmitrybelyakov:
Hey @richgilbank !

Yeah, that worked, thank you!. But it still behaves weirdly. Can you please tell me what's the purpose of this functionality? Because:

  * There is no way to get subtotal when initialising the cart this way
  * As soon as we add anything else on top of what was passed into 'constructor' all the items from before disappear
  * If we have to fetch every product in the cart to display images - will we hit API rate limit pretty quickly? (Say we have 5 users each having 10 items in the cart).

richgilbank:
[...]

PR #71, opened 2016-04-04
Fix line item documentation for cart creation

richgilbank:
Per https://github.com/Shopify/js-buy-sdk/issues/69, line items currently have to be passed in as an array of objects under the `line_items` key to `createCart`: https://github.com/Shopify/js-buy-sdk/blob/master/tests/integration/shop-client-cart-test.js#L20

### DIFF
--- a/guide/index.md
+++ b/guide/index.md
@@ -56,11 +56,11 @@ shopClient.createCart().then(function (cart) {
 });
 ```
 
-Variants can be passed in during initalization to create a non-empty cart with those variants.
+Line items can be passed in during initalization to create a non-empty cart with those variants.
 
 ```js
 var cart;
-shopClient.createCart({id: 123, quantity: 1}).then(function (cart) {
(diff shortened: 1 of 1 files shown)


richgilbank:
Bump 😄

tessalt:
LGTM

minasmart:
This won't totally work with local storage cart. There's no way for the cart to tie the line item back to a product, find the price, or the image. For this to work we'd need to send in variant classes merged with a quantity off of the product model. That might work right now but I haven't tested it. Line items are a mix of product and variant data so it may take some extra work to get this going.

richgilbank:
Hmm, yeah I thought it would work if a variant ID and a quantity were explicitly passed, or at least it would be able to generate a checkout permalink since those are the two pieces needed for it. Doesn't seem to do anything with it unless it has a full variant in the line item though. Will either change this section to pass in a full variant or remove for now.

PR #72, opened 2016-04-04
Docs fixes / additions

richgilbank:
There are several changes and fixes included here:
* Remove the yuidoc scripts from package.json, since they're only ever used from the gh-pages branch (as far as I know)
* Rename hash => object since JS doesn't have hashes.
* Make generic methods (`fetch`, `fetchAll`, `fetchQuery`, etc) private so they're not visible by default
* Fix various inconsistencies in naming
* Improve docs for functions that are wrappers of the private methods ^
* Add annotations for `fetchCart()` which were missing
* Document queries available to `fetchQueryProducts` and `fetchQueryCollections`

### DIFF
--- a/src/shop-client.js
+++ b/src/shop-client.js
@@ -96,7 +96,7 @@ const ShopClient = CoreObject.extend({
    * ```
    *
    * @method fetchAll
-   * @public
+   * @private
    * @param {String} type The pluralized name of the type, in lower case.
    * @return {Promise|Array} a promise resolving with an array of `type`
    */
@@ -118,7 +118,7 @@ const ShopClient = CoreObject.extend({
    * ```
    *
    * @method fetch
-   * @public
+   * @private
    * @param {String} type The pluralized name of the type, in lower case.
    * @param {String|Number} id a unique identifier
    * @return {Promise|BaseModel} a promise resolving with a single instance of
@@ -142,7 +142,7 @@ const ShopClient = CoreObject.extend({
    * ```
    *
    * @method fetchQuery
-   * @public
+   * @private
    * @param {String} type The pluralized name of the type, in lower case.
    * @param {Object} query a query sent to the api server.
    * @return {Promise|Array} a promise resolving with an array of `type`.
@@ -156,7 +156,7 @@ const ShopClient = CoreObject.extend({
   },
 
   /**
-   * Create an instance of `type`, optionally including `attrs`.
+   * Create an instance of `type`, optionally including `modelAttrs`.
    *
    * ```javascript
    * client.create('carts', { line_items: [ ... ] }).then(cart => {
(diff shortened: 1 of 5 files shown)


richgilbank:
Pls review @tessalt @harismahmood89 :)

tessalt:
having the yuidoc server on master is useful for when you're writing inline docs (making sure you've got the syntax right etc)

harismahmood89:
LGTM

src/shop-client.js
```diff
@@ -271,15 +270,19 @@ const ShopClient = CoreObject.extend({
   },
 
   /**
-    * Convenience wrapper for {{#crossLink "ShopClient/update:method"}}
-    * {{/crossLink}}. Equivalent to:
+    * Updates an existing {{#crossLink "CartModel"}}CartModel{{/crossLink}} instance and saves it.
```
tessalt: so this method no longer makes an API call, correct? We should be consistent about indicating what 'saves' means.

src/shop-client.js
```diff
@@ -271,15 +270,19 @@ const ShopClient = CoreObject.extend({
...
+    * client.createCart().then(cart => {
+    *   cart.lineItems = [
+    *     // ...
+    *   ];
+    *   client.updateCart(cart);
+    * });
     * ```
     *
     * @param {Object}[attrs={}] attributes representing the internal state of the cart to be persisted to the server.
-    * @method createCart
```
tessalt: I believe `persisted to the server` is no longer accurate

src/shop-client.js
richgilbank: Good point, will change to `...instance and persists it to LocalStorage`

src/shop-client.js
```diff
@@ -142,7 +142,7 @@ const ShopClient = CoreObject.extend({
    * ```
    *
    * @method fetchQuery
-   * @public
+   * @private
```
tessalt: I'm not sure if it's cool to mark public methods as private for the purpose of hiding them in the docs... not sure what best practice is here. @m-ux might have an opinion? It makes it clearer in the API docs but makes it confusing in the source.

src/shop-client.js
richgilbank: Good eye!

src/shop-client.js
richgilbank: Hm, yeah. I was originally going to go with `protected` here, but the [JSDoc ref](http://usejsdoc.org/tags-protected.html) suggested it shouldn't be used outside of a particular module. However, the [YUIdoc ref](http://yui.github.io/yuidoc/syntax/#protected) says it's for subclassing which seems a little closer to the use here. Thoughts?

tessalt: looking at the docs it does say "methods that _shouldn't_ be used externally" so I guess it works either way.

richgilbank:
Cool, I'm going to keep the server task but remove the other one, then :ship: :)

src/shop-client.js
m-ux: in Slate we're marking private methods as private for the purpose of hiding them from the docs... But IMO the 2 should go hand in hand.  Same with `protected` -- because both generally has meaning in terms of OOP, I'm hesitant to use either for presentation purposes. 

Is there a specific reason you guys want to hide a public method from the docs (maybe an alternative is simply don't doc-block this method?)?

richgilbank: Good question. Essentially, they're the methods that provide the behaviour leveraged by higher-level functions. The main concern is that they introduce a large amount of complexity to the documentation, because it introduces two ways of doing the same thing. They're not really private, _or_ protected, they just provide the behaviour behind the API we want to expose. Not documenting these might be a good approach as well. Hmmm... 🤔

PR #74, opened 2016-04-05
update examples/index.md to include quantity parameter

Brianshayne:
Changed example for cart permalink to include parameter listed in reference http://shopify.github.io/js-buy-sdk/api/classes/ProductVariantModel.html#method-checkoutUrl

@minasmart @richgilbank for review

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -40,7 +40,7 @@ client.fetchProduct('your-product-id').then(function(product) {
   "<img class='product__image' src='" + product.selectedVariantImage.src + "' >" +
   "<h2 class='product__title'>" + product.title + "</h2>" +
   "<a class='product__buy' href='" +
-    product.selectedVariant.checkoutUrl({quantity: 1}) +
+    product.selectedVariant.checkoutUrl([quantity = 1]) +
   "'>Buy Now!</a>";
 
   $('#product-1').html(html);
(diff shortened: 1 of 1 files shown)


minasmart:
:+1:

examples/index.md
```diff
@@ -40,7 +40,7 @@ client.fetchProduct('your-product-id').then(function(product) {
   "<img class='product__image' src='" + product.selectedVariantImage.src + "' >" +
   "<h2 class='product__title'>" + product.title + "</h2>" +
   "<a class='product__buy' href='" +
-    product.selectedVariant.checkoutUrl({quantity: 1}) +
+    product.selectedVariant.checkoutUrl([quantity = 1]) +
```
richgilbank: Since this is an example, this should probably just be 
```javascript
product.selectedVariant.checkoutUrl()
```
Or, if we want to be explicit about the quantity:
```javascript
product.selectedVariant.checkoutUrl(1)
```

richgilbank:
I'd be inclined to go with the former option (`checkoutUrl()`). The `[quantity = 1]` would make sense in reference docs where you're saying what the default value would be, but in an example case like this, it should be just using those methods directly.

richgilbank:
:+1:

Brianshayne:
Updated, and included the explicit quantity since we had it in the previous example and define it explicitly elsewhere.

PR #76, opened 2016-04-05
Change reference entry point to ShopifyBuy

richgilbank:
We should be jumping into the entry point for the SDK by default instead of the Cart Model. 


tessalt:
makes sense.

harismahmood89:
:sheep: :it:

PR #80, opened 2016-04-11
[FIX] Copy script tag to clipboard

richgilbank:
@harismahmood89 mind running through this and taking over it if there's anything outstanding?


harismahmood89:
@richgilbank yep! just top-hatted. looks good!

PR #82, opened 2016-04-12
Cart Example Enhancements

harismahmood89:
- Click away handlers
- Close cart when `esc` key is pressed
- Open/close cart function cleanups
- Retrieve previously selected item on cart close
- Clean up handlers

Please review @xto @richgilbank

### DIFF
--- a/examples/cart/index.js
+++ b/examples/cart/index.js
@@ -24,7 +24,7 @@ $(function() {
     });
   }
 
-
+  var previousFocusItem;
 
 
   /* Fetch product and init
@@ -52,7 +52,7 @@ $(function() {
 
     attachCheckoutButtonListeners();
 
-    closeCart();
+    bindEventListeners();
 
   });
 
@@ -119,13 +119,15 @@ $(function() {
       event.preventDefault();
       var id = product.selectedVariant.id;
       addVariantToCart(product.selectedVariant, 1);
+      setPreviousFocusItem(this);
+      $('#checkout').focus();
     });
   }
 
   /* Increase product variant quantity in cart
   ============================================================ */
   function attachQuantityIncrementListeners(product) {
-    $('.cart').on('click', '.quantity-increment', function() {
+    $('.cart').on('click', '.quantity-increment', function(evt) {
       var variantId = parseInt($(this).attr('data-variant-id'), 10);
       var variant = product.variants.filter(function (variant) {
         return (variant.id === variantId);
@@ -135,13 +137,14 @@ $(function() {
       $(this).attr('disabled', 'disabled');
 
       addVariantToCart(variant, 1);
+      evt.stopPropagation();
     });
   }
 
   /* Decrease product variant quantity in cart
   ============================================================ */
   function attachQuantityDecrementListeners(product) {
-    $('.cart').on('click', '.quantity-decrement', function() {
+    $('.cart').on('click', '.quantity-decrement', function(evt) {
       var variantId = parseInt($(this).attr('data-variant-id'), 10);
       var variant = product.variants.filter(function (variant) {
         return (variant.id === variantId);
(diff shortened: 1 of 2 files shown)


examples/cart/index.js
```diff
@@ -158,13 +160,40 @@ $(function() {
...
   ============================================================ */
   function closeCart() {
+    $('.cart').removeClass('js-active');
+    $('.overlay').removeClass('js-active');
+  }
+
+  /* Close Cart Handler
+  ============================================================ */
+  function closeCartHandler() {
     $('.cart .btn--close').click(function () {
```
minasmart: `.click(closeCart)`

examples/cart/index.js
```diff
@@ -158,13 +160,40 @@ $(function() {
...
+      closeCart();
+    });
+  }
+
+  /* Key Press Handler
+  ============================================================ */
+  function keyPressHandler() {
+    $(document).keydown(function(e){
+       var code = e.keyCode || e.which;
+       if (code === 27) {
```
minasmart: Make this a constant somewhere. Like: `var ESCAPE_KEYCODE === 27`, then `if (code === ESCAPE_KEYCODE)`. Try to stay away from magic numbers.

minasmart:
Two tiny things, otherwise 🚢

minasmart:
:shipit:

examples/cart/index.js
```diff
@@ -52,7 +52,9 @@ $(function() {
 
     attachCheckoutButtonListeners();
 
-    closeCart();
+    closeCartHandler();
+    overlayHandler();
+    keyPressHandler();
```
richgilbank: Any reason to keep these separated out as their own functions? Seems like it might be a little more simple to have them squashed into a single `bindEventListeners()` or similar call, particularly given the number of them on this page.

examples/cart/index.js
```diff
@@ -158,13 +160,37 @@ $(function() {
...
-    $('.cart .btn--close').click(function () {
-      $('.cart').removeClass('js-active');
+    $('.cart').removeClass('js-active');
+    $('.overlay').removeClass('js-active');
+  }
+
+  /* Close Cart Handler
+  ============================================================ */
+  function closeCartHandler() {
+    $('.cart .btn--close').click(closeCart);
```
richgilbank: I think `.on('click', ...` is preferred here, though it looks like it's no longer explicitly stated in the JS style guide. @lemonmade, was that removed intentionally or just lost in the migration?

examples/cart/index.js
```diff
@@ -158,13 +160,37 @@ $(function() {
...
+  ============================================================ */
+  function overlayHandler() {
+    $('.overlay').click(closeCart);
+  }
+
+  /* Key Press Handler
+  ============================================================ */
+  function keyPressHandler() {
+    var ESCAPE_KEYCODE = 27
+    $(document).keydown(function(e){
```
richgilbank: Same `.on('keydown', ...` comment as above ^

examples/cart/styles/_overlay.scss
```diff
@@ -0,0 +1,12 @@
+.overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: none;
+
+  &.js-active {
+    display: block;
```
richgilbank: Worth adding a transition here to fade in the overlay?

examples/cart/index.js
```diff
@@ -158,13 +160,37 @@ $(function() {
...
+       var code = e.keyCode || e.which;
+       if (code === ESCAPE_KEYCODE) {
```
richgilbank: Should be able to remove the above line, since `e.which` is normalized by jQuery.
```diff
- if(code === ESCAPE_KEYCODE) {
+ if(evt.which === ESCAPE_KEYCODE) {
```

examples/cart/index.js
richgilbank: Also, `evt` is preferred, [per the style guide](https://github.com/Shopify/javascript#2.6) :)

examples/cart/index.js
lemonmade: on is still preferred, it's just moved to the jQuery-specific guide: https://github.com/Shopify/javascript/blob/master/jquery/README.md

examples/cart/index.js
harismahmood89: Good idea, cleaning up listeners

examples/cart/styles/_overlay.scss
harismahmood89: the overlay is transparent, just there to act as a big event handler for clicking away from the cart and providing a place for customizing the overlay if wanted

harismahmood89:
@richgilbank all updates made!

examples/cart/styles/_overlay.scss
richgilbank: In that case, we probably shouldn't be using an overlay at all - we should be listening globally for clicks, then ensuring they aren't within the cart element. Otherwise, people would be clicking on elements on the page to no effect. As an example,
```javascript
$(document).on('click', function(evt) {
  if(!$(evt.target).closest('.cart').length) {
    closeCart();
  }
});
```

harismahmood89:
okay so using `stopPropagation` to handle event handling race conditions and exploring a better solution for the cart in a separate PR. 

Review eyes please @minasmart @tessalt

PR #84, opened 2016-04-13
Focus on checkout button when adding product to cart

harismahmood89:
- Auto focus on `#checkout` button in cart when triggering an "add to cart" event.

Please review @xto @richgilbank


minasmart:
👍

richgilbank:
Since the cart's essentially functioning as a modal window, focus should probably be returned to its previous state if the modal (/cart) is closed, i.e by pressing escape (much like the "Get Started" dialog on shopify.com)

harismahmood89:
Good call, @richgilbank

harismahmood89:
Adding this functionality to #82

PR #85, opened 2016-04-13
Complete fix for copy to clipboard

richgilbank:
Turns out I didn't commit everything I had for https://github.com/Shopify/js-buy-sdk/pull/80 - it's currently not functional in iOS, and we shouldn't be showing the copy button if it's never going to work for them. This hides the button if copying to clipboard isn't supported at all. 

@harismahmood89 @tessalt

### DIFF
--- a/assets/scripts/docs.js
+++ b/assets/scripts/docs.js
@@ -40,16 +40,21 @@ $(function() {
     generateApiClassNavItems();
   }
 
-  var clipboard = new Clipboard('[data-clipboard-text]');
-  clipboard.on('success', function(e) {
-    var previousText = $(e.trigger).text();
-    $(e.trigger).text('Copied!');
-    setTimeout(function() {
-      $(e.trigger).text(previousText);
-    }, 2000);
-  });
-  clipboard.on('error', function(e) {
-    $(e.trigger).text('Press ⌘-C now to copy');
-  });
+  if(document.queryCommandSupported('copy')) {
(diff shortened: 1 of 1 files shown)


assets/scripts/docs.js
```diff
@@ -40,16 +40,21 @@ $(function() {
...
+    var clipboard = new Clipboard('[data-clipboard-text]');
+    clipboard.on('success', function(e) {
+      var previousText = $(e.trigger).text();
+      $(e.trigger).text('Copied!');
+      setTimeout(function() {
+        $(e.trigger).text(previousText);
+      }, 2000);
+    });
+    clipboard.on('error', function(e) {
+      $(e.trigger).text('Press ⌘-C now to copy');
```
harismahmood89: Does using the symbol have any support issues? We can use `&#8984;` [Reference](http://www.interactiveoctopus.com/blog/index.php/2008/01/display-the-apple-command-symbol-in-html/)

minasmart: Should be OK, so long as we make the script tag like `<script src="..." charset="utf-8">`

minasmart:
👍

assets/scripts/docs.js
richgilbank: #TIL 
Definitely thought `charset="utf-8"` was assumed by default but [nope](https://github.com/atom/language-html/issues/53#issuecomment-91430302)

PR #89, opened 2016-04-15
[FEATURE] Support cross domain linking with google analytics

minasmart:
Autodetect google analytics, and fetch the linker param. Append the linker param
to a cart's `checkoutUrl`. This URL expires in two minutes, so best usage is to
fetch the URL when the `Checkout` button is clicked.

For reference: https://developers.google.com/analytics/devguides/collection/analyticsjs/linker

### DIFF
--- a/examples/cart/index.html
+++ b/examples/cart/index.html
@@ -89,5 +89,16 @@
     </div>
   </script>
 
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+    ga('create', 'UA-49178120-19', 'auto');
+    ga('send', 'pageview');
+
+  </script>
+
 </body>
 </html>
--- a/src/models/cart-model.js
+++ b/src/models/cart-model.js
@@ -74,7 +74,20 @@ const CartModel = BaseModel.extend({
       return `${item.variant_id}:${item.quantity}`;
     });
 
-    const query = `api_key=${config.apiKey}`;
+    let query = `api_key=${config.apiKey}`;
+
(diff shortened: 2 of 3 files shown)


minasmart:
@tessalt and @richgilbank could you take a look?

tessalt:
:+1:

minasmart:
@lreeves wanna take a look around?

lreeves:
🚢 💯

examples/cart/index.html
```diff
@@ -89,5 +89,16 @@ <h2 class="cart-title">Your cart</h2>
     </div>
   </script>
 
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+    ga('create', 'UA-49178120-19', 'auto');
```
richgilbank: Where's this ID from?

minasmart: Discussed IRL, it came from our analytics team and is used only here.

craigmichaelmartin:
Thank you for this great SDK! 😄  I'm sorry to comment on an issue long-closed.

Does the final checkout page append this value as a query param to the thank_you page url? 

If it helps, I ask because my checkout pages have no google analytics script, meaning that either:
- I am missing something in the configuration of my buy button to instruct the checkout page to have a google analytics script (with the `allowLinker` field set to true in the creation of the tracker). Without a google analytics script on the checkout page, as is the case for me now, this `_ga` query parameter would have no effect.
- Some internal Shopify checkout javascript would have to store this value and apply it to the `thank_you` page url when the time comes (or put it in the cookie how google expects). Shopify gives me access to add a google analytics script at this point (the `thank_you` page), and so the internal javascript would bridge the `checkout` pages gap (where I seem to have no ability to add GA). However, if this is the method, what happens if checkout takes longer than 2 minutes (the time until the linkerParam value expires)?
- Something else I don't understand 🙃 

This would be easy enough for me to test, but I have lots of pipelines set up on order creation and would prefer not to trip those 😬

PR #95, opened 2016-04-20
[FIX] Separate dev dependencies from run-time dependencies

minasmart:
This also fixes a couple bugs in running this thing as a node module.

Additionally, the package gets built on pre-publish


package.json
```diff
@@ -16,7 +16,12 @@
   "author": "Shopify Inc.",
   "license": "MIT",
   "dependencies": {
+    "node-fetch": "1.3.3"
+  },
+  "devDependencies": {
```
richgilbank: 😻

src/metal/global.js
```diff
@@ -0,0 +1,11 @@
...
+
+let globalNamespace;
+
+if (typeof global === 'undefined') {
+  globalNamespace = window;
+} else {
+  globalNamespace = global;
+}
+
+export default globalNamespace;
```
richgilbank: Could this be simplified to `export default global || window`?

richgilbank:
@tessalt @lreeves Can one of you two take a peek at this as well?

src/metal/global.js
minasmart: Nah. If global is undefined that will throw. Also, no logic in an export statement.

lreeves:
It looks good as far as I can tell!

PR #101, opened 2016-05-02
Cart Rework

harismahmood89:
Built on from #82.

Any changes to the line items would call a complete re-render of the cart DOM causing a couple of issues. This has been reworked to update the cart DOM's state instead of re-rendering it entirely.

- `renderCartItem()` renders a single line item. This function is called when adding a single item, or recursively when rendering the entire cart on page load.
- `updateVariantInCart()` is used to update an already existing item in the cart. Updates the `quantity` and `price` values.
- `variantToCartHandler()` determines whether to call the update or render function.
- `evt.stopPropagation()` is no longer required in the increment and decrement functions to prevent bubbling.
- Small tweaks for esc key handler.

### DIFF
--- a/examples/cart/index.html
+++ b/examples/cart/index.html
@@ -16,11 +16,11 @@
     <h2 class="variant-title"></h2>
     <h2 class="variant-price"></h2>
     <div class="variant-selectors"></div>
-    <button class="buy-button js-preventCartListener">Add To Cart</button>
+    <button class="buy-button js-prevent-cart-listener">Add To Cart</button>
   </div>
 
   <!-- .cart-tab start -->
-  <button class="btn btn--cart-tab js-preventCartListener">
+  <button class="btn btn--cart-tab js-prevent-cart-listener">
     <span class="btn__counter"></span>
     <svg xmlns="http://www.w3.org/2000/svg" class="icon-cart icon-cart--side" viewBox="0 0 25 25" enable-background="new 0 0 25 25"><g fill="#fff"><path d="M24.6 3.6c-.3-.4-.8-.6-1.3-.6h-18.4l-.1-.5c-.3-1.5-1.7-1.5-2.5-1.5h-1.3c-.6 0-1 .4-1 1s.4 1 1 1h1.8l3 13.6c.2 1.2 1.3 2.4 2.5 2.4h12.7c.6 0 1-.4 1-1s-.4-1-1-1h-12.7c-.2 0-.5-.4-.6-.8l-.2-1.2h12.6c1.3 0 2.3-1.4 2.5-2.4l2.4-7.4v-.2c.1-.5-.1-1-.4-1.4zm-4 8.5v.2c-.1.3-.4.8-.5.8h-13l-1.8-8.1h17.6l-2.3 7.1z"></path><circle cx="9" cy="22" r="2"></circle><circle cx="19" cy="22" r="2"></circle></g></svg>
   </button>
--- a/examples/cart/index.html
+++ b/examples/cart/index.html
@@ -16,8 +16,16 @@
     <h2 class="variant-title"></h2>
     <h2 class="variant-price"></h2>
     <div class="variant-selectors"></div>
-    <button class="buy-button btn">Add To Cart</button>
+    <button class="buy-button js-preventCartListener">Add To Cart</button>
   </div>
+
+  <!-- .cart-tab start -->
+  <button class="btn btn--cart-tab js-preventCartListener">
+    <span class="btn__counter"></span>
+    <svg xmlns="http://www.w3.org/2000/svg" class="icon-cart icon-cart--side" viewBox="0 0 25 25" enable-background="new 0 0 25 25"><g fill="#fff"><path d="M24.6 3.6c-.3-.4-.8-.6-1.3-.6h-18.4l-.1-.5c-.3-1.5-1.7-1.5-2.5-1.5h-1.3c-.6 0-1 .4-1 1s.4 1 1 1h1.8l3 13.6c.2 1.2 1.3 2.4 2.5 2.4h12.7c.6 0 1-.4 1-1s-.4-1-1-1h-12.7c-.2 0-.5-.4-.6-.8l-.2-1.2h12.6c1.3 0 2.3-1.4 2.5-2.4l2.4-7.4v-.2c.1-.5-.1-1-.4-1.4zm-4 8.5v.2c-.1.3-.4.8-.5.8h-13l-1.8-8.1h17.6l-2.3 7.1z"></path><circle cx="9" cy="22" r="2"></circle><circle cx="19" cy="22" r="2"></circle></g></svg>
+  </button>
+  <!-- .cart-tab end -->
+
 
   <!-- .cart begin -->
   <div class="cart">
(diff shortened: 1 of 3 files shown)


harismahmood89:
@richgilbank @minasmart @tessalt Please review when you have a moment - would like to get this out ASAP

examples/cart/index.js
```diff
@@ -118,39 +118,78 @@ $(function() {
...
+      setPreviousFocusItem(this);
+      $('#checkout').focus();
     });
   }
 
   /* Increase product variant quantity in cart
   ============================================================ */
   function attachQuantityIncrementListeners(product) {
-    $('.cart').on('click', '.quantity-increment', function() {
+    var $DomNode = $(this);
```
tessalt: Confused about the value of `this` here... wouldn't it just be the global object?

examples/cart/index.js
```diff
@@ -118,39 +118,78 @@ $(function() {
...
+
+    $('.cart').on('click', '.quantity-increment', function(evt) {
       var variantId = parseInt($(this).attr('data-variant-id'), 10);
       var variant = product.variants.filter(function (variant) {
         return (variant.id === variantId);
       })[0];
 
-      $(this).closest('.cart-item').addClass('js-working');
-      $(this).attr('disabled', 'disabled');
+      $DomNode.closest('.cart-item').addClass('js-working');
```
tessalt: yeah I'm pretty sure this is just getting `window`... how is this working?

examples/cart/index.js
```diff
@@ -163,62 +202,111 @@ $(function() {
...
+
+  /* Determine action for variant adding/updating/removing
+  ============================================================ */
+  function variantToCartHandler(variant, quantity) {
+    openCart();
+    var foundInCart = cart.lineItems.filter(function (item) {
+      return (item.variant_id === variant.id);
     });
+
+    if (foundInCart.length > 0) {
```
tessalt: `0` is falsy in JS, you can just do `if (foundInCart.length)`

examples/cart/index.js
```diff
@@ -163,62 +202,111 @@ $(function() {
...
+  }
 
   /* Add 'quantity' amount of product 'variant' to cart
   ============================================================ */
   function addVariantToCart(variant, quantity) {
     openCart();
 
     cart.addVariants({ variant: variant, quantity: quantity }).then(function() {
-      renderCartItems();
+      var $cartItem = renderCartItem(cart.lineItems[cart.lineItems.length - 1]);
```
tessalt: I'm confused about why we're rendering the last item in `cart.lineItems` and not a specific item?

tessalt:
couple questions but looking good. I'm sure we could figure out a way to get rid of the `setTimeout` hacks but I get that we want to 🚢

examples/cart/index.js
harismahmood89: `addVariantToCart` adds a variant that currently doesn't exist in the cart, and is therefore the last item in the list. So we can just grab the last item and render that instead of doing a search for it first.

minasmart: I wouldn't rely on this behaviour. If we swap it out for a server-persisted version, there's no guarantee around ordering. The server can re-order, and insert items as it needs to.

harismahmood89:
@tessalt @minasmart thanks for the review! made some changes based on your feedback. Let me know if there's anything else.

tessalt:
👍

examples/cart/index.js
```diff
@@ -163,62 +182,114 @@ $(function() {
...
   function closeCart() {
-    $('.cart .btn--close').click(function () {
-      $('.cart').removeClass('js-active');
+    $('.cart').removeClass('js-active');
+    $('.overlay').removeClass('js-active');
+  }
+
+  /* Determine action for variant adding/updating/removing
+  ============================================================ */
+  function variantToCartHandler(variant, quantity) {
```
minasmart: This name isn't very clear to me.

examples/cart/index.js
```diff
@@ -118,39 +118,58 @@ $(function() {
...
+    $('.cart .btn--close').on('click', closeCart);
+
+    $(document).on('click', function(evt) {
+      if((!$(evt.target).closest('.cart').length) && (!$(evt.target).closest('.js-preventCartListener').length)) {
+        closeCart();
+      }
+    });
 
-      addVariantToCart(variant, -1);
+    var ESCAPE_KEYCODE = 27
```
minasmart: This should live up near the top where you define any other globals, constants, whatevs.

examples/cart/index.html
```diff
@@ -16,9 +16,17 @@ <h1 class="product-title"></h1>
     <h2 class="variant-title"></h2>
     <h2 class="variant-price"></h2>
     <div class="variant-selectors"></div>
-    <button class="buy-button btn">Add To Cart</button>
+    <button class="buy-button js-preventCartListener">Add To Cart</button>
```
richgilbank: Should probably be consistently kebob-case

examples/cart/index.js
```diff
@@ -52,7 +52,7 @@ $(function() {
 
     attachCheckoutButtonListeners();
 
-    closeCart();
+    bindEventListeners();
```
richgilbank: This should probably include the other listeners as well. Right now there's a bunch of specific functions being called, then one generic that is named like it would include all of them, but doesn't 😕

harismahmood89:
@tessalt @minasmart wanna check this out again? lots of fixes - would suggest some 🎩 -ing too if you can

examples/cart/index.js
```diff
@@ -114,44 +157,80 @@ $(function() {
...
-      event.preventDefault();
-      var id = product.selectedVariant.id;
-      addVariantToCart(product.selectedVariant, 1);
-    });
+  function buyButtonClickHandler(evt) {
+    evt.preventDefault();
+    var id = product.selectedVariant.id;
+    var quantity;
+    var cartLineItem = findCartItemByVariantId(id);
+    if (cartLineItem) {
```
tessalt: you can squish this into one line with a ternary operator

examples/cart/index.js
```diff
@@ -114,44 +157,80 @@ $(function() {
...
-  /* Decrease product variant quantity in cart
+  /* Increase product variant quantity in cart
   ============================================================ */
-  function attachQuantityDecrementListeners(product) {
-    $('.cart').on('click', '.quantity-decrement', function() {
-      var variantId = parseInt($(this).attr('data-variant-id'), 10);
-      var variant = product.variants.filter(function (variant) {
-        return (variant.id === variantId);
-      })[0];
+  function incrementQuantity(evt) {
```
tessalt: is `evt` being used?

examples/cart/index.js
```diff
@@ -114,44 +157,80 @@ $(function() {
...
+    var variant = product.variants.filter(function (variant) {
+      return (variant.id === variantId);
+    })[0];
+    var quantity;
+    var cartLineItem = findCartItemByVariantId(variant.id);
+    if (cartLineItem) {
+      quantity = cartLineItem.quantity + 1;
+      updateVariantInCart(cartLineItem, quantity);
+    }
+  }
```
tessalt: I can think of a couple ways we could combine the increment/decrement/input quantity methods. One way would be to have an `updateQuantity` function which accepts a function as a parameter. That function would accept the current value and return the new value. Then in your `updateQuantity` function you could say 
```
var quantity = fn(currentQuantity)
```

and then call it with like
```
updateQuantity(function (current) {
   return current + 1;
});
```

Or: 

```
var addOne = function (current) {return current + 1}
var incrementQuantity = updateQuantity.bind(addOne);
```

harismahmood89:
updates made based on feedback. let me know if there's anything else @tessalt @minasmart

examples/cart/index.js
```diff
@@ -114,44 +163,78 @@ $(function() {
...
-      addVariantToCart(variant, -1);
-    });
+  /* Debounce taken from _.js
+  ============================================================ */
+  function debounce(func, wait, immediate) {
+  	var timeout;
+  	return function() {
+  		var context = this, args = arguments;
+  		var later = function() {
+  			timeout = null;
```
minasmart: What happened to indents here?

examples/cart/index.js
```diff
@@ -114,44 +163,78 @@ $(function() {
...
-      $(this).closest('.cart-item').addClass('js-working');
-      $(this).attr('disabled', 'disabled');
+  function updateQuantity(fn, variantId) {
+    var variant = product.variants.filter(function (variant) {
+      return (variant.id === variantId);
+    })[0];
+    var quantity;
+    var cartLineItem = findCartItemByVariantId(variant.id);
+    if (cartLineItem) {
+      quantity = fn(cartLineItem.quantity);
```
tessalt: 😍

tessalt:
lookin much better

PR #104, opened 2016-05-03
[FIX] line item serialization

harismahmood89:
- Add a line item class so deserialized line items have getters
- Required testing

Most paired with @minasmart 

Please review: @richgilbank @tessalt @lreeves

PR #107, opened 2016-05-04
[FIX] Cart line item serialization and deserialization

minasmart:
This is a redo of #104. Now with the added bonus of not dumping `CartLineItemModel` classes to local storage, and a test to make sure we don't intermingle internal state and decorated state.


minasmart:
@harismahmood89 I fixed it! And added a new test. It was a tiny change, and a longer test. Have a look!

Also @tessalt if you want to have a look, that'd be stellar.

harismahmood89:
🚀

tessalt:
👍

PR #109, opened 2016-05-12
[FIX] Proxy line item IDs properly

minasmart:
This didn't create any issues with how the SDK works, but for anyone
consuming the SDK, they wouldn't have had unique IDs on enumerables that
are persisted. This could have made it more difficult to implement
client side work using line items.

Please review @harismahmood89 and @lreeves

### DIFF
--- a/src/models/cart-line-item-model.js
+++ b/src/models/cart-line-item-model.js
@@ -7,7 +7,7 @@ const CartLineItem = BaseModel.extend({
   },
 
   get id() {
-    return this[GUID_KEY];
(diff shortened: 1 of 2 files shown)


lreeves:
Nice, curious how you two noticed this one. 👍  but only because it's been GPG signed.

minasmart:
@lreeves: @harismahmood89 was trying to check for the existence of a line item by using the for a uniqueness check. This showed him that there were no IDs, when they should have been there.

harismahmood89:
👍

PR #111, opened 2016-05-13
Fixed doc on `cart.addVariants` to reflect the correct implementation…

### LINKED ISSUE #110: Confusing example for addVariants in documentation
The example for calling the addVariants method shows an object with an ID and a quantity.
screen shot 2016-05-13 at 12 30 46 pm
However, the parameters for the item object state that you must pass a 'variant' and a 'quantity'.
screen shot 2016-05-13 at 12 30 56 pm
Maybe it's just me, but I found this really confusing. At first I thought I was supposed to pass the ID of the product. Then I thought it was the variant ID. Eventually I realized that I needed to pass the selectedVariant object, so now I have it working.

yomexzo:
Fixed doc on `cart.addVariants` to reflect the correct implementation.
Fixes #110

@minasmart, @harismahmood89

### DIFF
--- a/src/models/cart-model.js
+++ b/src/models/cart-model.js
@@ -98,7 +98,7 @@ const CartModel = BaseModel.extend({
   /**
     * Add items to cart. Updates cart's `lineItems`
     * ```javascript
-    * cart.addVariants({id: 123, quantity: 1}).then(cart => {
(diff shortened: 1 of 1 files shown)


harismahmood89:
👍

Issue #10, opened 2016-03-10
Image sizes


richgilbank:
Images returned from the API are the raw image size uploaded to Shopify, i.e. 
`https://cdn.shopify.com/s/files/1/1015/1595/products/blackhelmet_store.png?v=1446233645`

We should be able to return these in the various sizes Shopify produces, which requires adding a suffix to the file name matching [these sizes](https://docs.shopify.com/themes/liquid/filters/url-filters#size-parameters). 

i.e. instead of `product.selectedVariantImage.src`, maybe `product.selectedVariantImage('thumb')`?

@minasmart @tessalt @harismahmood89

minasmart:
Closing in favour of #51, but keeping it around for the links and details. :smile:

PR #114, opened 2016-05-17
Support selection of different variant scaled images.

yomexzo:
### Details
`ProductVariantModel.scaledImage()` is implemented, which is now called/used by `ProductVariantModel.image` getter.

The currently proposed schema to get the variant scaled images looks like the below
```
  product: {
    variant: {
      image: {
        imageVariants: [
          {
            name: 'thumb',
            dimensions: '50x50',
            src: 'https://cdn.shopify.com/image-two_thumb.jpg'
          }
        ]
      }
    }
```
### Assumptions
- This PR assumes current data source does not have an `image` key on the product variant data. In which case, the `ProductVariantModel.image` getter would work like it currently does.
- It also assumes that the `master` image variant will always be included when the data source has `image.imageVariants`

references - #10, #51, Shopify/shopify#71114

### Suggestion
To access the scaled images, we'd use `product.variant.image.imageVariants[]`. Since we know each variant would only have one image with different scales, i think we can change it to `product.variant.images[]`

@minasmart @lreeves

### DIFF
--- a/src/models/product-variant-model.js
+++ b/src/models/product-variant-model.js
@@ -91,11 +91,37 @@ const ProductVariantModel = BaseModel.extend({
     * @type {Object}
   */
   get image() {
-    const id = this.id;
-    const images = this.attrs.product.images;
+    return this.scaledImage();
+  },
 
-    const primaryImage = images[0];
-    const variantImage = images.filter(image => {
+  /**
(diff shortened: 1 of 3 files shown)


yomexzo:
Going to re-implement - wrong assumptions.

PR #119, opened 2016-05-19
fix removeLineItem and add tests to ensure proper lineItems update

harismahmood89:
Please review @minasmart @lreeves @tessalt

### DIFF
--- a/src/models/cart-model.js
+++ b/src/models/cart-model.js
@@ -211,7 +211,9 @@ const CartModel = BaseModel.extend({
     const newLength = newLineItems.length;
 
     if (newLength < oldLength) {
-      this.attrs.line_items = newLineItems;
+      this.attrs.line_items = newLineItems.map(item => {
+        return item.attrs;
+      });
 
       return this.updateModel();
     }
--- a/tests/unit/models/cart-model-test.js
+++ b/tests/unit/models/cart-model-test.js
@@ -350,3 +350,34 @@ test('it keeps line_item attrs hashes, and doesn\'t inject classes into internal
     done();
   });
 });
+
+
+test('it doesn\'t pollute "attrs.line_items" with "CartLineItem" class instances on #removeLineItem', function (assert) {
+  assert.expect(4);
+
+  const done = assert.async();
+
+  const quantity = 1;
+  const variantOne = singleProductFixture.product_listing.variants[0];
+  const variantTwo = singleProductFixture.product_listing.variants[1];
+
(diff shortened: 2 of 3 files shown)


lreeves:
👍

src/models/cart-model.js
```diff
@@ -211,7 +211,9 @@ const CartModel = BaseModel.extend({
     const newLength = newLineItems.length;
 
     if (newLength < oldLength) {
-      this.attrs.line_items = newLineItems;
+      this.attrs.line_items = newLineItems.map(item => {
```
tessalt: this can be a one-liner 😄 
`newLineItems.map(item => item.attrs)`

tests/unit/models/cart-model-test.js
```diff
@@ -350,3 +350,34 @@ test('it keeps line_item attrs hashes, and doesn\'t inject classes into internal
     done();
   });
 });
+
+
+test('it updates line_item attrs correctly when removing a line item', function (assert) {
+  assert.expect();
```
tessalt: i believe you want to pass the number of assertions as an argument to `expect` here

tests/unit/models/cart-model-test.js
```diff
@@ -350,3 +350,34 @@ test('it keeps line_item attrs hashes, and doesn\'t inject classes into internal
     done();
   });
 });
+
+
+test('it updates line_item attrs correctly when removing a line item', function (assert) {
```
minasmart: What is `correctly`?

Specifically: `it doesn't pollute "attrs.line_items" with "CartLineItem" class instances on #removeLineItem`

src/models/cart-model.js
harismahmood89: linter doesn't like that @minasmart

minasmart: Two questions: Is it valid without a return? And also: always braces.

minasmart: Er.... question and a statement.

minasmart:
I'm fine with this, pending @tessalt's concern.

src/models/cart-model.js
tessalt: single-line fat arrows without the braces implicitly return. I like the "as-needed" option for this rule but I'll check the shopify eslint config http://eslint.org/docs/rules/arrow-body-style

minasmart: I'm not a fan of 'as needed' because then the moment you need to grow the method, you need to start decorating it with more stuff. I'm a big fan of consistent, explicit conventions.

tessalt:
@minasmart cool until we adopt shopify/javascript eslintrc...

PR #121, opened 2016-05-20
[FEATURE] Report SDK info (version, variant) in requests

minasmart:
Also adds version info to the payload dynamically, as a comment, and through string interpolation.

Addresses a portion of #51.

### DIFF
--- a/build-lib/util/versioner.js
+++ b/build-lib/util/versioner.js
@@ -65,17 +65,20 @@ Versioner.prototype.versionFiles = function () {
     });
 
     return fileAcc.concat(list);
-  }, []).filter(fileName => {
-    return fileName.match(/^.+\.js$/);
-  });
+  }, []);
 
   files.forEach(fileName => {
     const inputBuffer = fs.readFileSync(path.join(fileName));
+    let outputBuffer;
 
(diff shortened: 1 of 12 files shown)


minasmart:
CC: @richgilbank 

👀 : @tessalt and @harismahmood89

build-lib/util/versioner.js
```diff
@@ -0,0 +1,84 @@
...
+  const versionString = `v${this.constructor.VERSION}-${this.constructor.HEAD_COMMIT}`;
+
+  const files = this.inputPaths.reduce((fileAcc, dir) => {
+    const list = fs.readdirSync(dir).map(fileName => {
+      return path.join(dir, fileName);
+    });
+
+    return fileAcc.concat(list);
+  }, []).filter(fileName => {
+    return fileName.match(/^.+\.js$/);
```
yomexzo: @minasmart, filtering of files to only `.js` files means the `outputPath` would only contain `.js` files. I think we should write all files to the output but version only the `.js` files. Otherwise, the consuming script would have to `mergeTrees` and use the `overwrite` config to ensure the other files are preserved.

tessalt:
it lgtm but i don't know anything about broccoli

PR #124, opened 2016-05-24
Remove node 0.10 from ci

minasmart:
Since we build on prepublish, no need to test all of the build rigging on CI. We can add this back in once we have proper node tests that install the module and consume it.


minasmart:
@tessalt @harismahmood89 please have a look!

tessalt:
👍

harismahmood89:
👍

PR #125, opened 2016-05-24
Add lineItemCount to CartModel

yomexzo:
This is part of the `v0.2.0` punch list on #51 

cc @tessalt , @minasmart , @harismahmood89


minasmart:
per #51, what you have as `totalQuantity` should be what `lineItemCount` returns. Also: test please!

yomexzo:
oh yes. I tested @minasmart 

`cart.lineItems.length` is `cart.lineItemCount` while getting a sum of all quantities per line item is `cart.totalQuantity` and i replaced accordingly in the example.

My understanding is a `lineItem` can have any quantity. But the example is displaying the sum of all quantities on the counter.

yomexzo:
The changes have been made. cc @minasmart

yomexzo:
@harismahmood89, @minasmart 🚪

examples/cart/index.js
```diff
@@ -240,11 +240,8 @@ $(function() {
   /* Update cart tab button
   ============================================================ */
   function updateCartTabButton() {
-    if (cart.lineItems.length > 0) {
-      var totalItems = cart.lineItems.reduce(function(total, item) {
-        return total + item.quantity;
-      }, 0);
-      $('.btn--cart-tab .btn__counter').html(totalItems);
+    if (cart.lineItemCount > 0) {
+      $('.btn--cart-tab .btn__counter').html(cart.totalQuantity);
```
minasmart: This change does not reflect your updates.

yomexzo: oh no! good catch 😄

tests/unit/models/cart-model-test.js
```diff
@@ -82,6 +84,7 @@ test('it creates a line item when you add a variant', function (assert) {
   model.addVariants({ variant, quantity }).then(cart => {
     assert.equal(cart, model, 'it should be the same model, with updated attrs');
     assert.equal(cart.lineItems.length, 2);
+    assert.equal(cart.lineItemCount, 3);
```
minasmart: This should live in its own test. It shouldn't be interjected into all these other tests. We test length, because we know that's the output of adding a variant.

lineItemCount is a unique key, built on top cart. It should have a unique test instead of coupling to all of these.

yomexzo:
@minasmart, just updated repo with `master` and made some changes.

kindly help review ... cc @tessalt @richgilbank 

Thanks.

src/models/cart-model.js
```diff
@@ -52,6 +52,17 @@ const CartModel = BaseModel.extend({
   },
 
   /**
+    * Gets the sum of quantity of each line item
```
minasmart: Pedantic, but I think this should be `Gets the sum quantity of each line item`

minasmart:
One tiny thing, otherwise :shipit: 

@tessalt or @harismahmood89 can you 👍  👎 ?

src/models/cart-model.js
tessalt: let's be pedantic, I think there's a lot of risk of confusion here.

tests/unit/models/cart-model-test.js
```diff
@@ -93,6 +95,25 @@ test('it creates a line item when you add a variant', function (assert) {
   });
 });
 
+test('it returns correct lineItemCount', function (assert) {
+  assert.expect();
```
tessalt: I think you're expecting `2` correct?

yomexzo: I think `done()` does the work here and is ideal when say the `success` callback has multiple `asserts` while the `error` callback has only one.

But yes, in this scenario, clearly two. You think it should be specified?

tessalt: we're doing that in the rest of the tests.

src/models/cart-model.js
yomexzo: Absolutely! `Gets the sum quantity of each line item` is confusing to me tho.
The word `lineItemCount` itself confuses me too 😄 . But i understand there's an existing shared understanding.

I'd remove the `of` since that makes sense to two people already 😄

tests/unit/models/cart-model-test.js
yomexzo: added 😄

minasmart: Yeah, expect should always specify a number. If assertions aren't run, we need to know.

richgilbank:
@yomexzo Can you update the docs site? I believe this is in the 0.1.7 release, but [not yet published](http://shopify.github.io/js-buy-sdk/api/classes/CartModel.html)

PR #127, opened 2016-05-26
Catered for when no image is set on product variants and product has no image listed

yomexzo:
`ProductVariant.image` would return an object with `src` set to `https://widgets.shopifyapps.com/assets/no-image.svg` and the rest of the keys set to `null`.

Developers can use `image.id` to determine if the "no image" image is being returned and can optionally swap with theirs.

However, `ProductVariant.imageVariants` returns `[]` in this scenario since we do not have different sizes for that image.


yomexzo:
changed the api approach to avoid breaking current usage of the SDK. A `noImageURI` was added to the `CartModel` instead which can be optionally used by developers.

cc @minasmart , @harismahmood89 , @lreeves

examples/cart/index.js
```diff
@@ -97,7 +97,9 @@ $(function() {
   /* Update product image based on selected variant
   ============================================================ */
   function updateVariantImage(image) {
-    $('#buy-button-1 .variant-image').attr('src', image.src);
+    var src = typeof image === 'undefined' ? cart.noImageURI : image.src;
```
minasmart: This will break if we switch it up to null. Why not just `src = (image) ? image.src : IMAGE_URI`

src/models/cart-model.js
```diff
@@ -96,6 +96,16 @@ const CartModel = BaseModel.extend({
   },
 
   /**
+    * Returns URI to an image that can be used as default image when a variant does
+    * not have an image. i.e when `variant.image` yields `undefined`
+    * @property noImageURI
+    * @type {String}
+  */
+  get noImageURI() {
+    return 'https://widgets.shopifyapps.com/assets/no-image.svg';
```
minasmart: So, this image is for products. It should be exported by the product module, not a getter on the cart. In globals mode, it should be an attribute on the global (`shopify.js`). It's also a constant, so it should be all caps. Not a getter.

src/models/product-variant-model.js
```diff
@@ -109,6 +109,11 @@ const ProductVariantModel = BaseModel.extend({
     * @type {Array}
   */
   get imageVariants() {
+    const image = this.image;
+
+    if (typeof image === 'undefined') {
```
minasmart: Just use `if (!image)`. Don't bind to an explicit falsey type. Just let it be falsey.

tests/unit/models/cart-model-test.js
```diff
@@ -381,3 +381,9 @@ test('it doesn\'t pollute "attrs.line_items" with "CartLineItem" class instances
   });
 
 });
+
+test('it returns Shopify admin\'s no image URI', function (assert) {
+  assert.expect(1);
+
+  assert.equal(model.noImageURI, 'https://widgets.shopifyapps.com/assets/no-image.svg');
```
minasmart: This key on the cart should go away.

tests/unit/models/product-variant-test.js
```diff
@@ -135,6 +135,12 @@ test('it returns the image variants for the variant', function (assert) {
   ];
 
   assert.deepEqual(model.imageVariants, expectedImageVariants);
+
+  const images = model.attrs.product.images;
```
minasmart: This is a distinct test.

yomexzo:
thanks @minasmart, changes have been implemented.

cc @tessalt, @harismahmood89, @lreeves

tests/unit/models/product-model-test.js
```diff
@@ -95,3 +95,9 @@ test('it attaches a reference to the config on variants', function (assert) {
     assert.equal(variant.config, config);
   });
 });
+
+test('it returns Shopify admin\'s no image URI', function (assert) {
+  assert.expect(1);
+
+  assert.equal(NO_IMAGE_URI, 'https://widgets.shopifyapps.com/assets/no-image.svg');
+});
```
minasmart: I would just test that this a string of non-zero length. I don't think we should be validating what the actual URI is.

minasmart:
Please rebase and squash so this can be merged in 😄

yomexzo:
Part of the punch list at #51

PR #131, opened 2016-06-06
Run tests in Node

yomexzo:
This is work in progress.
- The `node-lib` directory is now generated in all environments. It used to be generated only in production environment.
- Test code is not transpiled into commonjs and included in `node-lib` now
- an entry/bootstrap script is created to create an environment to run QUnit tests.

Test currently fails because of things like `localStorage not being defined`. But it's the first phase.

cc @tessalt, @minasmart, @harismahmood89


yomexzo:
This PR is pretty now ready for review. Since parts of it have been broken down into separate PRs, they'd become cleaner and more more readable when #160, #163 and #164 have been merged.

👀 @tessalt @m-ux @minasmart

yomexzo:
### Some context
Tests are written with QUnit.
The tests were written mostly with browsers in mind. e.g `pretender` library, used to mock api calls would only work in browsers.
We currently run tests via `testem` in `PhantomJS`.

This means we have no solid way of verifying that the library works as expected in NodeJS env.
This PR implements a bootstrap file (`scripts/util/test-in-node.js`) for node that makes tests work within the node environment.
It churns out logs in `tap` format and is hooked up to `testem`. So you would notice extra lines prefixed with `NodeQunit` when you run `npm run test`.

yomexzo:
👀 @tanema  😄

tanema:
Looks good to me, 😎

minasmart:
Take a look over here if you could: https://github.com/Shopify/js-buy-sdk/pull/171

I think some of the work I did there, if it could be merged in, would simplify a bit of the changes in the build portion of this PR

build-lib/util/mkdirp.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+const fs = require('fs');
+const path = require('path');
+
+function mkdirp(fullPath) {
+
```
minasmart: 🔥  no blank line immediately inside the function.

build-lib/util/recursive-read-dir.js
```diff
@@ -0,0 +1,21 @@
+"use strict";
+
+const fs = require('fs');
+const path = require('path');
+
+function recursiveReadDir(dir) {
+
```
minasmart: 🔥

package.json
```diff
@@ -46,6 +46,7 @@
     "eslint-config-nightmare-mode": "2.3.0",
     "estraverse-fb": "1.3.1",
     "fake-xml-http-request": "1.4.0",
+    "fetch-pretender": "^1.5.0",
```
minasmart: follow the convention of using exact versions please.

src/isomorphic-fetch.js
```diff
@@ -1,9 +1,16 @@
...
 import isNodeLikeEnvironment from './metal/is-node-like-environment';
 
 const fetch = global.fetch;
 
 if (!fetch && isNodeLikeEnvironment()) {
-  global.fetch = global.require('node-fetch');
+  /* this indirection is needed because babel throws errors when
+   * transpiling require('node-fetch') using `amd` plugin with babel6
+   */
+  const localRequire = require;
```
minasmart: This should go away with a rebase.

testem.json
```diff
@@ -3,8 +3,15 @@
   "cwd": ".dist-test",
   "test_page": "index.html?hidepassed",
   "disable_watching": true,
+  "launchers": {
+    "NodeQUnit": {
+      "command": "node scripts/util/test-in-node.js",
```
minasmart: Scripts that are run directly shouldn't live in util. They should be runnable scripts.

Util is for things that are required by other scripts.

tests/unit/adapters/listings-adapter-test.js
```diff
@@ -1,7 +1,7 @@
 import { module, test } from 'qunit';
 import ListingsAdapter from 'shopify-buy/adapters/listings-adapter';
 import version from 'shopify-buy/version';
-import Promise from 'promise';
+import 'shopify-buy/isomorphic-btoa';
```
minasmart: This should go in your boot up script.

build-lib/testing.js
```diff
@@ -51,14 +51,20 @@ function testsTree(pathConfig) {
   ]);
 
   const testModules = babelTranspiler(mergeTrees([testJs, linterTests]), babelConfig(`${pkg.name}/tests`, 'amd'));
+  const testsCommon = babelTranspiler(testJs, babelConfig(pkg.name, 'commonjs'));
```
minasmart: Why not just make this the same as the line above, but in commonjs instead of amd? Why are you excluding the name spacing for test modules? Why skip the linter?

minasmart: Also, if we now have a common tree and an amd tree, maybe use a similar naming convention for both, like in the other bits of build-lib? (suggestion: `testAmdTree` and `testCommonTree`)

build-lib/util/mkdirp.js
yomexzo: 👍 i'd be able to remove the file entirely once #172 is merged

build-lib/util/recursive-read-dir.js
yomexzo: same with this. re: #172

package.json
yomexzo: again 😞

src/isomorphic-fetch.js
yomexzo: that would be possible with #164 which is not merged yet.

testem.json
yomexzo: so the thing is, it can not be run directly. It depends on execution of `scripts/test` which uses `testem`. That's why i moved it into util. Wasn't sure where exactly it fits

yomexzo: it depends on `test` because it needs the `.dist-test` directory to have been built

yomexzo:
Rebased this branch on top of master only to resolve conflicts. Need to go back to how it was implemented since lots of changes have been made since this PR was made

osamie:
@minasmart @yomexzo is this still a work in progress or can we close this PR?

PR #132, opened 2016-06-06
[UPGRADE] Use babel 6

minasmart:
This upgrades the version of babel that we're using through the broccoli-babel-transpiler.
It opens us up do doing things like using the transform-runtime/core-js.


minasmart:
CC: @yomexzo 

Please review @tessalt and @harismahmood89

tessalt:
lgtm as far as I can tell.

yomexzo:
👍  @minasmart

PR #135, opened 2016-06-07
[FIX] Handle variants that don't exist

minasmart:
Instead of blowing up, guard for a variant that doesn't exist and return null variants and null images


minasmart:
Please review @harismahmood89 and @tessalt

harismahmood89:
awesome 👍

tessalt:
👍

PR #138, opened 2016-06-13
update version in package.json

tessalt:
I thought 1.6 was the latest but looks like that went out to npm in april. Bumping to 0.1.7. 

@Shopify/buy-button

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "shopify-buy",
-  "version": "0.1.6",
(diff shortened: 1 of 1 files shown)


tessalt:
hm CI is failing presumably because the git tag is at 0.1.6 and package.json is at 0.1.7. I kind of think that the tagging shouldn't happen until the package.json version is updated in master... anyone have an opinion on this?

PR #139, opened 2016-06-13
expose `Product.description`

### LINKED ISSUE #103: Expose product descriptions under `bodyHtml`
We left this out from the product model. We should be proxying it so it's available as `product.bodyHtml` so developers don't have to rely on internal state (`product.attrs.body_html`).

Original issue: #100

yomexzo:
Fixes #103

👀  @tessalt, @minasmart, @harismahmood89

### DIFF
--- a/src/models/product-model.js
+++ b/src/models/product-model.js
@@ -34,11 +34,11 @@ const ProductModel = BaseModel.extend({
   },
 
   /**
-    * Product bodyHtml
-    * @property bodyHtml
+    * Product descriptionHtml
+    * @property descriptionHtml
     * @type {String}
   */
-  get bodyHtml() {
+  get descriptionHtml() {
     return this.attrs.body_html;
   },
 
--- a/src/models/product-model.js
+++ b/src/models/product-model.js
@@ -34,7 +34,7 @@ const ProductModel = BaseModel.extend({
   },
 
   /**
-    * Product description
+    * Product description. The exposes the `body_html` property on the listings API
     * @property description
     * @type {String}
   */
--- a/src/models/product-model.js
+++ b/src/models/product-model.js
@@ -34,11 +34,11 @@ const ProductModel = BaseModel.extend({
   },
 
   /**
-    * Product descriptionHtml
-    * @property descriptionHtml
+    * Product description
+    * @property description
     * @type {String}
   */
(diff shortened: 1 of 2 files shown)


minasmart:
Looks good to me. I think the CI failure is an intermittent thing around dependency fetching. Let me know if it persists.

richgilbank:
Bit of a nitpick, but should this maybe be `bodyHTML`? 
https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

Alternatively, should this be called `description`? It's called body_html in the api response, but I think description might be more...descriptive.

minasmart:
@richgilbank the decision to call it `bodyHtml` was driven by simply camelizing the `body_html` attr. `bodyHTML` is definitely more accurate, but represents a different pattern of proxying our API. I'm really not sure what the better way to go is. Thoughts?

richgilbank:
How so? We're not camelizing the attrs automatically, so wouldn't it be our own perogative to name it in a way that makes sense for SDK consumers?

yoadsn:
I would not know how strict you try to remain with the naming. But clearly the decision to name this property of the product as "body_html" is rooted in the past of the Shopify product as a solely web based platform.
The "Html" part is accurate enough (It is Html content after all) but the "body" part is not, body has no meaning, for example on native mobile apps which might use the API to get the product details. (not this SDK though).

It should have been descriptionHTML IMHO, but pretty sure you would not go with that but rather maintain backward (logical) compatibility.

Regarding the casing, I think you should leave it 'Html' since there are also 'Url' here and there which should have been 'URL'. So, if camel casing acronym is the way, let it be.

minasmart:
@yoadsn good points! Thank you!

After some internal discussion: If we want to optimize for the best developer experience, let's not bike shed over how to most closely mirror our internal data model. We call this field `description` in the admin, so let's call it `description` here.

Thanks all!

tessalt:
👍

minasmart:
I know this is a trivial addition in terms of CI, but could you rebase just to get the new stuff that makes CI work again?

minasmart:
Otherwise 👍

yomexzo:
Everything is up-to-date here. I will try a rebuild without cache.

PR #140, opened 2016-06-13
read version from package.json

tessalt:
rather than calling `exec` to get git tag, get version number for cdn scripts from package.json. I think this is a better approach but I'm also slightly suspicious about what my original motives were for not doing this the first time around. @minasmart @yomexzo @richgilbank

### DIFF
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -4,7 +4,7 @@ var AWS = require('aws-sdk');
 var fs = require('fs');
 var path = require('path');
 var mime = require('mime-types');
-var exec = require('child_process').exec;
+var tag = require('../package.json').version;
 var DEFAULT_CONTENT_TYPE = 'application/octet-stream';
 var S3_DIR = 'js-buy-sdk';
 var DEFAULT_CACHE_CONTROL = "public, max-age=86400";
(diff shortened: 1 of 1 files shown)


scripts/deploy
```diff
@@ -31,9 +31,15 @@ var Uploader = function () {
   this.secrets = new Secrets();
   this.secrets.load();
   this.s3 = new AWS.S3();
+  this.tag = this.getPkgVersion();
   this.distAssets = fs.readdirSync(DIST_DIR);
 };
 
+Uploader.prototype.getPkgVersion = function () {
+  var pkg = fs.readFileSync('./package.json')
```
minasmart: You can actually do `require('./package.json')` I think. Which will remove the need to manually parse json on the next line.

scripts/deploy
```diff
@@ -53,24 +59,12 @@ Uploader.prototype.uploadFile = function(localPath, s3Path) {
...
-  }.bind(this));
-};
-
-Uploader.prototype.getCurrentTag = function (cb) {
-  exec('git describe --always --tag --abbrev=0', (err, stdout) => {
-    if (err) {
-      console.error(err);
-      cb();
+  this.distAssets.forEach(function (file) {
+    if (file.indexOf(FILENAME_FILTER) > -1) {
```
minasmart: use `file.contains(FILENAME_FILTER)`

scripts/deploy
tessalt: nice

richgilbank:
:+1:

minasmart:
:shipit:

PR #141, opened 2016-06-14
add deploying docs

tessalt:
I'm not sure what the best order to do things in is. The way I've written it, you will have to push the version bump commit to master, which isn't great. However, if you run `npm version` from the branch, then the npm package and master will not be at the same sha, since master will have a merge commit. Thoughts? 

@minasmart @richgilbank @harismahmood89

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -1,0 +1,11 @@
+# Deploying JS Buy SDK
+
+1. Merge your branch into master
+2. Run `npm version [patch|minor|major]` which will do the following:
+  * write new version to package.json
+  * create a commit with a commit message matching the version number
+  * create a new tag matching the version number
+3. Push the new commit and tags to master with `git push origin master --tags`
+4. Run `npm update` to update the package on npm
+5. Create a release on Github. Include changelog in description
+6. Deploy built scripts to s3 via [shipit](https://shipit.shopify.io/shopify/js-buy-sdk/production)


richgilbank:
I think this makes sense for now. One thing though is that we should decide if it's worth having a [CHANGELOG.md](https://github.com/Shopify/js-buy-sdk/blob/master/CHANGELOG.md) since the release notes should summarize things anyway. Thoughts? My inclination would be to :fire: it down.

tessalt:
agreed, notes for releases should be the source of truth.

minasmart:
Disagree. Release notes are a fairly hard to discover feature of GitHub that lives entirely outside the got repo and have no aggregation mechanism. While I agree that we've done a poor job of maintaining the change log, it actually lives in the repo and provides aggregated, discoverable, data. I have a script that should get built into our release processes, but I feel like punting everything to release notes makes it easier to hide us not documenting releases. The change log makes our obligation more apparent.

tessalt:
Fair enough. I think ideally having both if we had a script to keep them in sync would be ideal, since i do really like the "download zip" feature of gh releases.

minasmart:
The script I have uses the github API to pull in release notes, then appends them to the doc. It's in ruby, but I could rewrite it, or we could just roll with it. Whatever you think is best.

tessalt:
ruby doesn't bother me.

minasmart:
For reference: https://gist.github.com/minasmart/7a531198e9559ed11a330e23caa642bd

minasmart:
It doesn't bother me, it's just a second language dependency for the repo. However, if it's only shopifolk who are able to delpoy, it shouldn't be that big a deal.

minasmart:
Also, to be clear, I don't think we shouldn't use releases. I just think we should use both, and mirror appropriately. 👍

yoadsn:
@tessalt A thought, If you would to rebase your branch on top of master and only then run `npm version` on it, your (immediate) merge onto master would be a fast-forward which should maintain the sha. (Assuming not commits to master between rebase and merge, which is a reasonable assumption for version closing step).

Did I miss something?

minasmart:
@yoadsn I feel at that point we're blessing a PR as the next version, when it really represents a point in time on master. Functionally it's the same, but workflow-wise I think it's much easier to run these changes from master.

yoadsn:
@minasmart Totally agree, it's a workflow derived decision. I worked on projects where dev branch sets the next version, and others where it was master. Just wanted to point out the sha problem is solvable with FF.

PR #142, opened 2016-06-14
Checkout example

yomexzo:
This example showcases how to manage the checkout process such that the cart can be cleared.

Reflects what @minasmart referred to in #98 

👀 @harismahmood89, @minasmart, @tessalt


tessalt:
seems legit.

examples/checkout/styles/_buttons.scss
```diff
@@ -0,0 +1,24 @@
...
+  overflow: hidden;
+  line-height: 20px;
+  -moz-appearance: none;
+  -webkit-appearance: none;
+
+  &:hover,
+  &:focus {
+    background: $color-primary-hover;
+  }
+}
```
minasmart: Make sure your editor is set up to end files with an EOL character.

Also @harismahmood89 can we have your FED eyes?

minasmart:
@yomexzo this looks great!

Could you add a readme indicating the intent of this example? We're illustrating how to monitor the checkout page and take action. Ideally, developers should be able to discover that from a doc.

Thanks!

examples/checkout/styles/_initializers.scss
```diff
@@ -0,0 +1,16 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  // font-family: 'Muli', sans-serif;
```
harismahmood89: 🔥

examples/checkout/index.html
```diff
@@ -0,0 +1,36 @@
...
+  <div class="cart not-empty">
+    <img class="variant-image">
+    <h2><span class="line-item-count"></span> items in cart</h2>
+    <h2>Sub-total: <span class="sub-total"></span></h2>
+    <button class="btn checkout">Checkout</button>
+  </div>
+
+  <div class="cart empty">
+    <h2>Thanks for checking out!!</h2>
+  </div>
```
harismahmood89: Do we need two cart items to handle their various DOM states? Should be a cleaner way of doing this. Let me know if you'd like some help

examples/checkout/index.html
```diff
@@ -0,0 +1,36 @@
...
+
+  <div class="cart not-empty">
+    <img class="variant-image">
+    <h2><span class="line-item-count"></span> items in cart</h2>
+    <h2>Sub-total: <span class="sub-total"></span></h2>
+    <button class="btn checkout">Checkout</button>
+  </div>
+
+  <div class="cart empty">
+    <h2>Thanks for checking out!!</h2>
```
harismahmood89: Being nitpick-y, but drop the second `!`

examples/checkout/index.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  /* Update UI variables
+  ============================================================ */
+  function completeUIRendering() {
+    $('.cart.empty').toggle(!cart.lineItemCount);
+    $('.cart.not-empty').toggle(cart.lineItemCount !== 0);
+
+    $('.cart.not-empty .sub-total').text(formatAsMoney(cart.subtotal));
+    $('.cart.not-empty .line-item-count').text(cart.lineItemCount);
```
harismahmood89: `$('.cart.not-empty')` is used repeatedly, maybe set it as as a var then use `.find` if you need to find elements inside it

harismahmood89:
Made a few small comments, otherwise 👍

examples/checkout/index.js
```diff
@@ -0,0 +1,70 @@
...
+    myShopifyDomain: 'embeds',
+    appId: '6'
+  });
+
+  var product;
+  var cart;
+
+  client.createCart().then(function (newCart) {
+    cart = newCart;
+    return client.fetchProduct('3614436099');
```
richgilbank: Does this need to be chained after `createCart` like this? Could they be in parallel? Right now the cart is basically instantaneous, but in the future these will be separate requests. 
```javascript
var cartPromise = client.createCart();
var productPromise = client.fetchProduct('3614436099');
Promise.all([cartPromise, productPromise]).then(function(values) {
  cart = values[0];
  product = values[1];
  cart.addVariants(...
});
```

examples/checkout/index.js
```diff
@@ -0,0 +1,70 @@
...
+      completeUIRendering();
+
+      // redirect to the home page or to custom thank you page.
+    }
+  }
+
+  /* Format amount as currency
+  ============================================================ */
+  function formatAsMoney(amount) {
+    return '$' + parseFloat(amount, 10).toFixed(2).replace(/(\d)(?=(\d{3})+\.)/g, "$1,").toString();
```
richgilbank: Regex ✨ magic✨  - can you add a comment explaining the behaviour?

examples/checkout/styles/_cart.scss
```diff
@@ -0,0 +1,16 @@
+.cart {
+  max-width: 400px;
+  box-shadow: 0 0 100px 10px #eee;
```
richgilbank: To make it more reusable in the future, please use variables for all colors.

examples/checkout/index.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  var product;
+  var cart;
+
+  client.createCart().then(function (newCart) {
+    cart = newCart;
+    return client.fetchProduct('3614436099');
+  }).then(function (product) {
+    cart.addVariants({ variant: product.variants[0], quantity: 5 }, { variant: product.variants[1], quantity: 3 }).then(function() {
```
minasmart: let's break this up into a few lines. It's purty long.

examples/checkout/index.js
```diff
@@ -0,0 +1,70 @@
...
+      completeUIRendering();
+      bindEventListeners();
+    }).catch(function (errors) {
+      console.log('Fail');
```
minasmart: Be more verbose here. If it's intended for a developer to see, use `console.info`, and include an explanation before just dumping the trace.

examples/checkout/index.html
```diff
@@ -0,0 +1,36 @@
...
+  <link rel="stylesheet" href="index.css">
+  <script src="//code.jquery.com/jquery-2.2.1.min.js"></script>
+  <script src="/shopify-buy.polyfilled.globals.js"></script>
+  <script src="/examples/checkout/index.js"></script>
+</head>
+<body>
+
+  <div class="cart not-empty">
+    <img class="variant-image">
+    <h2><span class="line-item-count"></span> items in cart</h2>
```
richgilbank: Let's make this 0 so it doesn't just display `items in cart` while the product is loading.

examples/checkout/index.html
```diff
@@ -0,0 +1,36 @@
...
+    <h2>Sub-total: <span class="sub-total"></span></h2>
```
richgilbank: `$0.00`

examples/checkout/styles/_buttons.scss
yomexzo: ah. thanks.

examples/checkout/index.js
yomexzo: gotcha

examples/checkout/index.js
yomexzo: lol .. that wasn't me!! i copied it from the other example 😄 
I just took it as what it is ... formats money ... LOL

@harismahmood89 ... do you mind? 😸

harismahmood89: The regex is comma separating every 3 characters in the price amount to convert `123456789.00` to `123,456,789.00`. Example was found here: http://stackoverflow.com/a/16963988

yomexzo: thanks @harismahmood89!!

In my opinion, the function name speaks for itself. Also, the regex doesn't look complicated at all. I see a find and replace using lookforward regex.

If we must put some note, the stackoverflow link should be adequate IMO .... @richgilbank

minasmart: Regexes are just messy, hard to read, and hard to maintain in general. They're also not very declarative. The more declarative approach would be something like:

```javascript
parseFloat(amount, 10).toFixed(2)
                      .split('') // transform to array
                      .reverse() // reverse direction
                      .join('') // turn back into a string
                      .match(/.{1,3}/g) // match every three characters, plus whatever's left over
                      .map(segment => segment.split('').reverse().join('')) // reverse each individual segment
                      .reverse() // reverse the three character chunks
                      .join(',') // join with commas.
```
In this case, I would definitely add a comment describing the action of the regex, but I don't think the declarative non-regex approach is any easier to understand.

richgilbank: In that case, it would be good to extract the dollars/cents separator and thousands separator so they can be replaced easily for locales with commas and periods the other way around - i.e. 
```
function formatAsMoney(amount) {
  var thousandsSeparator = ',';
  var centsSeparator = '.';
  var parts  = parseFloat(amount, 10).toFixed(2).split('.');
  var dollars = parts[0].replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1' + thousands);
  return '$' + dollars + centsSeparator + parts[1];
}
```
Per: https://github.com/Shopify/widget-framework/blob/master/app/assets/javascripts/framework/vendor/money.js#L30

yomexzo: honestly, both seems complex and hard for me to read at a glance 😆 . Either i am just biased because I have a thing for regex or may be it's just me that thinks `/(\d)(?=(\d\d\d)+(?!\d))/g` is harder to read or may be i just don't get the point entirely 😄 .


How about we add the reference to where regex came from? Lots of context there. The question, the answer, etc.

richgilbank: I'm happy with using either regex, was just using the one we've used before. If the first one is more readable to you, that works (neither are particularly readable to me). My main concern is about being able to change the currency formatting easily.

yomexzo: Makes absolute sense @being able to change formatting. I'd use the `formatAsMoney` function you shared.

yomexzo:
@minasmart @richgilbank .. the changes have been made ... kindly take a second 👀

examples/cart/index.js
```diff
@@ -371,8 +371,18 @@ $(function() {
 
   /* Format amount as currency
   ============================================================ */
-  function formatAsMoney(amount) {
-    return '$' + parseFloat(amount, 10).toFixed(2).replace(/(\d)(?=(\d{3})+\.)/g, "$1,").toString();
+  function formatAsMoney(amount, currency, thousandSeparator, decimalSeparator, localeDecimalSeparator) {
+    amount = 1000.3
```
richgilbank: I think this was left in accidentally

examples/checkout/index.js
```diff
@@ -0,0 +1,91 @@
...
+    myShopifyDomain: 'embeds',
+    appId: '6'
+  });
+
+  var product;
+  var cart;
+
+  var cartPromise = client.createCart();
+  var productPromise = client.fetchProduct('3614436099');
+  RSVP.all([cartPromise, productPromise]).then(function (values) {
```
richgilbank: Shouldn't this be `Promise.all` instead? Surprised RSVP is exposed here at all, but it should be polyfilling `Promise.all` if not available natively.

examples/checkout/index.js
```diff
@@ -0,0 +1,91 @@
...
+      completeUIRendering();
+
+      // redirect to the home page or to custom thank you page.
+    }
+  }
+
+  /* Format amount as currency
+  ============================================================ */
+  function formatAsMoney(amount, currency, thousandSeparator, decimalSeparator, localeDecimalSeparator) {
+    amount = 1000.3
```
richgilbank: Same as above ^

examples/checkout/index.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
```
richgilbank: Please add `<meta name="viewport" content="width=device-width">` here for both examples. Just noticed the cart example currently isn't responsive.

examples/cart/index.js
yomexzo: oh no. definitely accidental

richgilbank:
Just noticed the cart example no longer has functioning styles for the button or select element, which both appear to be white: . 

Any idea what might have caused this? cc @harismahmood89

examples/checkout/index.html
yomexzo: okay

examples/checkout/index.js
yomexzo: indeed, `Promise.all` is ideal. Only that `RSVP` exposes it too. Changing it now.

yomexzo:
@richgilbank @harismahmood89, I fixed the button bit by changing the class `buy-button` to `btn` ... `buy-button` wasn't defined anywhere.

examples/checkout/index.js
minasmart: Yeah, try to be polyfill agnostic. I'd like to remove RSVP and run on core-js eventually.

yomexzo:
@harismahmood89, i did some investigation with @richgilbank, the box around the select box disappears when in Chrome Mobile view. I checked with Firefox mobile/responsive view and it looks perfect.

Have no idea what the problem is with the thing in Chrome :)

harismahmood89:
@yomexzo hmm that's strange, I'll take a look

yomexzo:
@richgilbank, the issue with the select dropdown not showing borders on mobile has now been sorted out. Styles from Shopify/shopify-buy-ui#11 were used. This means we have a consisted look for the dropdown across all devices. Thanks @harismahmood89.

cc @tessalt 

👀

yomexzo:
Added the `README` file and rebased.

harismahmood89:
looks good! 👍

tanema:
I 🎩  it and it looks good to go.

Issue #133, opened 2016-06-07
Safari private mode


jneree:
Hi,
It's been now 2 days I'm trying to make your JS Buy SDK work perfectly on my website www.soundbrenner.com/shop

I have a very annoying issue with Safari in private (incognito) mode: it's impossible to add a product in the cart.
After testing, I found that the client.createCart() function is not working in the SDK (the client variable remains undefined), again only in Safari private mode. Did anyone have the same issue?

website: www.soundbrenner.com/shop

Thanks a lot! 
Julien

minasmart:
We're working on a fix for this. Watch #143 for updates.

PR #143, opened 2016-06-16
store cart in memory if localStorage not available

tessalt:
Fixes https://github.com/Shopify/js-buy-sdk/issues/133

Safari in private browsing mode throws `QuotaExceededError: Dom exception 22` on `localStorage.setItem` (and returns `null` for `localStorage.getItem`). Carts should still work in private browsing mode, so fall back to storing carts in memory. This is pretty much the quickest/simplest implementation, not sure if it would be better to create a separate "adapter" instead and check for localStorage before assigning an adapter to the `carts` type. 

@minasmart @harismahmood89 @richgilbank


minasmart:
Based on our quick talk, let's re-work this by proxying localStorage, and handling this complexity in a `store` (or some other better named) class.

tessalt:
@minasmart updated with `store`. I didn't come up with a better name... 
I'm also a little unhappy with the tests now. But guess what, they pass!

tests/integration/shop-client-cart-test.js
```diff
@@ -49,9 +49,11 @@ test('it resolves with a new cart on ShopClient#create', function (assert) {
   delete cartAttrs[GUID_KEY];
 
   localStorage.setItem = function (key, value) {
-    assert.ok(key.match(/shopify-buy\.\d+\.\d+/));
-    assert.deepEqual(JSON.parse(value).cart.line_items, cartAttrs.line_items);
-    assert.equal(JSON.parse(value).cart.subtotal_price, cartAttrs.subtotal_price);
+    if (key !== '__storage_test__') {
```
tessalt: so yeah this sucks... the first call to `localStorage` is now the feature test, so these assertions fail the first time.

minasmart: maybe test local storage using a `{}` that you've called `setGUIDFor` on, and use `testObj[GUID_KEY]` in the try/catch above as the key? Then we're using the supported storage mechanism, but your test still functions right?

minasmart: That way, these tests could remain untounched.

src/store.js
```diff
@@ -0,0 +1,48 @@
...
+    if (this.localStorageAvailable) {
+      const stringValue = localStorage.getItem(key);
+
+      try {
+        return JSON.parse(stringValue);
+      } catch (e) {
+        return null;
+      }
+    } else {
+      return this.cache[key] || null;
```
richgilbank: I believe this would return `null` if the value is explicitly `false`. Unlikely to be an issue, but just in case...

richgilbank:
:+1:

src/store.js
tessalt: i think it could only ever be the string `"false"`

tests/integration/shop-client-cart-test.js
tessalt: @minasmart not sure quite what you mean, do you mean in the test as in shop-client-cart-test or test as in the feature test for localstorage?

minasmart: nevermind. I pulled it down and tried what I was suggesting. Your way is the better way here 👍

minasmart: Maybe in a later PR, we can make the store to use configurable, and pass in our fake local storage, but for now, this works.

nocean:
With the merge above it seems this was fixed, but I'm struggling to figure out how to roll it out in a live setting. I'm using the [latest release](http://sdks.shopifycdn.com/js-buy-sdk/v0/latest/shopify-buy.umd.polyfilled.min.js), but am still having problems with this. From the 'cart' example, pasted below, how would I modify this code to work?

```
/* Build new ShopifyBuy client
  var client = ShopifyBuy.buildClient({
    apiKey: 'bf081e860bc9dc1ce0654fdfbc20892d',
    domain: 'embeds.myshopify.com',
    appId: '6'
  });

  var product;
  var cart;
  var cartLineItemCount;
  if(localStorage.getItem('lastCartId')) {
    client.fetchCart(localStorage.getItem('lastCartId')).then(function(remoteCart) {
      cart = remoteCart;
      cartLineItemCount = cart.lineItems.length;
      renderCartItems();
    });
  } else {
    client.createCart().then(function (newCart) {
      cart = newCart;
      localStorage.setItem('lastCartId', cart.id);
      cartLineItemCount = 0;
    });
  }

  var previousFocusItem;
```

minasmart:
The best way to fetch the most recently used cart is with [`fetchRecentCart`](https://github.com/Shopify/js-buy-sdk/blob/master/src/shop-client.js#L418-L433)

However, in sandboxed environments like private browsing mode, this will always just return a new cart. 

If this doesn't work, can you let me know what version of the SDK you're using and how you're consuming it? (npm, CDN, etc)

Hope this helps!

nocean:
Thanks @minasmart. I'm using the CDN (the latest version, linked in my comment above).

I'm relying heavily on the examples provided in the repo -- any help rewriting the example code (from the repo, pasted above) to handle this appropriately is very much appreciated. 

Is there a reason the cart example doesn't use fetchRecentCart, but uses the conditional test on localStorage instead?

minasmart:
The example should be corrected. 

As for helping you rewrite: what's the desired outcome? Using the fetch recent method with no arguments and no conditional should work 100% of the time.

nocean:
Spent some time with this and was able to rewrite the example code to use fetchRecentCart and it's working beautifully -- no more error.

However, while I can now add products to the cart with Safari's Private Browsing on, the cart clears and starts fresh on every page load. This is different behaviour from Firefox's "Private Window" and Chrome's "Incognito Window".

Shouldn't the cart be stored in memory and available from page to page (at least while the browser remains open)?

minasmart:
Sadly there aren't any standards around private browsing. In Chrome and Firefox, local storage is available and persists for the duration of your private session. In Safari's private browsing mode, the local storage API is completely disabled. That means carts are stored in the JavaScript program's memory. Every page refresh clears that memory and starts over.

nocean:
Yeah, I can appreciate the differences between browsers. However, it's a frustrating limitation of the Buy SDK though, as Shopify.com sites can handle a traversing cart in a Safari private window. I presume it's done by storing the cart information in a cookie instead.

minasmart:
.com persists carts through sending the data to the server, and returning a token that's used in the URL. The browser doesn't have to be responsible for retaining anything. The next major version of the SDK will support this through a server side cart API. Sorry there's not a great solution for private browsing for now.

nocean:
Thanks @minasmart. Appreciate the help and the explanation. Looking forward to the next major version. Is there a place where I can follow that release?

minasmart:
This repo is the best place to follow along. Thanks for your interest in our SDK :)

Jens99:
Hi @minasmart Is there any update on when the next major version you refer to will be ready? Thanks!

minasmart:
No specific date. Working on a beta candidate presently though.

PR #144, opened 2016-06-16
[FIX] version and link node modules.

minasmart:
Circle caches dependencies from VM 0. This means that statically compiled
modules in node_modules from whatever node version is being tested on VM 0 get
distributed to every other node version when caches are distributed. This
creates a versioned `node_modules_X` directory, and symlinks `node_modules` to
it.


minasmart:
@tessalt @lreeves I think this fixes broken things. Mind having a look?

lreeves:
👍 🔢

tessalt:
👍

Issue #122, opened 2016-05-24
[docs] Switch between versions


richgilbank:
Currently the docs are independent of the lib version - we should be able to switch between versions and display the docs from each one.

tessalt:
hm that's probably going to be a significant update to the hacky way we're doing docs. @m-ux have you done any more investigation into documentation solutions?

m-ux:
only really know about jsdoc at the moment, but jsdoc generates to a subfolder based on the version number specified in your package.json

When you rev the version number & regenerate the docs, it will go in a new folder.  

One thing you can do (lodash docs used to do this, but are slightly different now) is have a dropdown to select version numbers that link to the appropriate code.  

There's also the `version`, `since` and `deprecated` properties, which could theoretically be used to more dynamically showcase stuff... but that could be a lot more complicated.

We've got a simple implementation of jsdoc outputting to raw json data which we'd like to use at some point in the future for a more robust solution - you'd likely need to approach the problem that way to take advantage of the version/since/deprecated properties.

lopugit:
do it, I'll help

mikkoh:
You can now generate docs for specific versions. It's documented here:
https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md#documentation

PR #149, opened 2016-06-28
Auto generation of docs and committing doc changes onto gh-pages branch

yomexzo:
So far, we have the API docs generated per tag and including master using the `yuidoc-lucid-theme`.
Next phase would be to use the modified version in `gh-pages` instead and possibly move context specific parts of the docs to each tag docs

kindly help review for any feedback / suggestion. @minasmart, @tessalt, @richgilbank


scripts/build
```diff
@@ -6,13 +6,22 @@ require('./util/work-from-root')();
 
 var cleanDirectory = require('./util/clean-directory');
 var buildProject = require('./util/build-project');
+var docBuilder = require('./util/build-doc')({ docsBranchName: 'gh-pages-6', commitDocs: true, rmSrcDir: true });
```
tessalt: what's with the `-6`

scripts/util/build-doc.js
```diff
@@ -0,0 +1,303 @@
...
+
+  var repo;
+
+  console.log(`\nChecking out ${self.options.docsBranchName} branch`);
+  NodeGit.Repository.open('.').then(function (_repo) {
+    repo = _repo;
+    console.log(`Getting ${self.options.docsBranchName}'s latest commit`);
+    return repo.getBranchCommit(self.options.docsBranchName);
+  }).then(function (commit) {
+    console.log(`Getting it's working tree`);
```
tessalt: `its`

scripts/util/build-doc.js
```diff
@@ -0,0 +1,303 @@
...
+  NodeGit.Repository.open('.').then(function (_repo) {
+    repo = _repo;
+
+    console.log(`Getting available references within the repo`);
+    return repo.getReferences(NodeGit.Reference.TYPE.OID);
+  }).then(function(references){
+    var treesDataPromises = [];
+
+    console.log(`Filtering them to include only tags and the ${self.options.masterBranchName} branch.`);
+    references.forEach(function(reference){
```
tessalt: you can probably use `map` here I think

tessalt:
couple comments but looks good. There are a lot of console logs... I think this is probably fine but if they're staying in I'd prefer if they were more specific. There are a few non-specific references written assuming the context of the previous log. Let's be explicit what we're talking about each time.

scripts/build
yomexzo: That has been removed now. I created multiple branches while performing tests locally.

scripts/util/build-doc.js
yomexzo: thanks @tessalt

yomexzo:
Thanks @tessalt for the feedback.

The changes have been made now and the whole work is not ready and no more "work in progress".

cc @minasmart, @m-ux, @richgilbank, @harismahmood89

package.json
```diff
@@ -62,6 +62,7 @@
     "wdio-mocha-framework": "^0.3.1",
     "webdriverio": "4.0.9",
     "whatwg-fetch": "0.10.1",
-    "yuidocjs": "0.10.0"
+    "yuidoc-lucid-theme": "^0.1.2",
```
minasmart: exact versions.

scripts/build
```diff
@@ -6,13 +6,26 @@ require('./util/work-from-root')();
 
 var cleanDirectory = require('./util/clean-directory');
 var buildProject = require('./util/build-project');
+var docBuilder = require('./util/build-doc')({
+	docsBranchName: 'gh-pages',
```
minasmart: indentation should only be two spaces here.

scripts/build
```diff
@@ -6,13 +6,26 @@ require('./util/work-from-root')();
 
 var cleanDirectory = require('./util/clean-directory');
 var buildProject = require('./util/build-project');
+var docBuilder = require('./util/build-doc')({
```
minasmart: maybe call the file `doc-builder` if you're calling the var `docBuilder`

scripts/build
```diff
@@ -6,13 +6,26 @@ require('./util/work-from-root')();
...
 buildProject(buildPath, function (buildStatus) {
-  process.exit(buildStatus);
+  if (buildStatus !== 0) {
+    return process.exit(buildStatus);
+  }
+
+  docBuilder.build(function () {
+    console.log('\nDone building');
+  })
+
```
minasmart: 🔥 empty line

scripts/util/build-doc.js
```diff
@@ -0,0 +1,293 @@
...
+var yuidoc = require('./yuidoc');
+var NodeGit = require('nodegit');
+var util = require('util');
+var recursiveRmdir = require('./recursive-rmdir');
+var recursiveCpdir = require('./recursive-cpdir');
+
+function buildTreePromise(commitPromise, name) {
+  return commitPromise.then(function (commit){
+    return commit.getTree();
+  }).then(function (tree){
```
minasmart: sticking the `then` here couples it to the body of the previous promise. Since this is scheduled for after `commit.getTree()`, maybe nest it there?

```javascript
commit.getTree().then(function (tree) { ... })

File: scripts/util/build-doc.js [hunk 9]
```diff
@@ -0,0 +1,293 @@
...
+      name: name,
+      tree: tree
+    }
+  });
+}
+
+/** This function checks out trees synchronously. 
+  * Important because attempting to checkout async causes errors
+  */
+function checkoutTrees(datum, rest, options, callback, paths) {
```
minasmart: Please use a better name than `datum`.

scripts/util/build-doc.js
```diff
@@ -0,0 +1,293 @@
...
+  var self = this;
+  if (fs.existsSync(self.options.docsDirName)) {
+    recursiveRmdir(self.options.docsDirName);
+  }
+
+  fs.mkdirSync(self.options.docsDirName)
+
+  var repo;
+
+  console.info(`\nChecking out ${self.options.docsBranchName} branch`);
```
minasmart: this sort of string interpolation won't work under node `v0.12`

scripts/util/build-doc.js
```diff
@@ -0,0 +1,293 @@
...
+  NodeGit.Repository.open('.').then(function (_repo) {
+    repo = _repo;
```
minasmart: This is an ugly way to capture the param. You should push this through with what's resolved, or properly nest your promise scopes.

scripts/util/build-doc.js
```diff
@@ -0,0 +1,293 @@
...
+          callback();
+        }
+      });
+    });
+  });
+}
+
+module.exports = function (options) {
+  return new DocBuilder(options);
+};
```
minasmart: turn on new line at EOF in your editor.

Also: There is a lot of coupled promise chaining here, which could be broken by a catch statement returning a value. If your then relies on the result of a previous promise, it should be chained off of that promise. There's also a lot of code that lives in every one of these scopes, which should probably be wrapped into functions.

scripts/util/yuidoc.js
```diff
@@ -0,0 +1,49 @@
+var Y = require('yuidocjs');
```
minasmart: `yuidoc`

shipit.yml
```diff
@@ -3,3 +3,4 @@ deploy:
     - npm install --no-progress
     - ejson decrypt -o config/secrets.json config/secrets.ejson
     - scripts/deploy
+    - git push origin gh-pages:gh-pages
```
minasmart: Don't include this just yet.

scripts/util/build-doc.js
yomexzo: the variable `repo` is used all over, won't passing it from one scope to another become messy? or what do you propose?

scripts/util/build-doc.js
yomexzo: fixed the editor. set it up initially but i probably messed it up again.

re: promise chaining, not sure i understand what you meant

yomexzo:
@minasmart, thanks for the feedback, the changes have been implemented as discussed. Kindly 👀

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,341 @@
...
+          if (self.options.commitDocs) {
+            self.commitAPIDocs(callback);
+          } else if (callback) {
+            callback();
+          }
+        });
+      });
+    });
+  });
+}
```
m-ux: This type of callback chain is something we want to avoid.  Can this be cleaned up with Promises instead?
I'd also suggest using `.bind()` instead of `var self = this`?

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,341 @@
...
+  recursiveCpdir(self.options.apiDirPath, apiTempDir);
+
+  NodeGit.Repository.open('.').then(function (repo) {
+
+    self.log('info', 'Getting repo\'s current staging area');
+    return repo.index().then(function (index) {
+      self.log('info', 'Getting ', self.options.docsBranchName, '\'s latest commit');
+      return repo.getBranchCommit(self.options.docsBranchName).then(function (docsBranchCommit) {
+        self.log('info', 'Getting the commit\'s tree');
+        return docsBranchCommit.getTree().then(function (tree) {
```
m-ux: these nested Promises are hard to read and reduce code clarity.  

There is some complex stuff going on here... Do any of these techniques help? https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,341 @@
...
+  self.log('info', '\nGenerating API docs');
+  yuidoc.generate({paths: paths, themeDir: self.options.themeDir}, function () {
+    self.log('info', 'Done generating API docs');
+    if (callback) {
+      callback();
+    }
+  });
+}
+
+DocBuilder.prototype.commitAPIDocs = function (callback) {
```
m-ux: these mixed in callbacks w/Promises add a lot of complexity.  Can these methods return promises instead?

scripts/util/doc-builder.js
minasmart: It's kind of ugly no matter what, but serially chaining them creates a lot of coupling. It also doesn't make the dependency chain clear. I need to read your link though 😉

scripts/util/doc-builder.js
minasmart: If you pull apart some of the promises into functions that return promises, it's going to be neater. 

And with `.bind`, if you use it on every function you pass to `.then`, `this` will persist down the chain.

yomexzo: okay, i knew that but i was thinking more in line of the end result. Look at the below for example.
I abstract the first inner block which returns `repo.getReferences` into a function because it uses `self`. Then we abstract `repo.getReferences`'s callback into it's own function so we can `.bind()`?

This is what i have pictured in my head and that's just one part of the codebase and even worse/deep dependencies on `self` exists. Except i'm missing something @minasmart 

```
 +  NodeGit.Repository.open('.').then(function (repo) {
 +
 +    self.log('info', 'Getting available references within the repo');
 +    return repo.getReferences(NodeGit.Reference.TYPE.OID).then(function(references){
 +      var treesDataPromises = [];
 +
 +      self.log('info', 'Filtering them to include only tags and the ', self.options.masterBranchName, ' branch.');
 +      references.forEach(function(reference){
 +        var target = reference.targetPeel() || reference.target();
 +        var name = reference.shorthand();
 +        var commitPromise;
 +
 +        if(reference.isTag()) {
 +          commitPromise = repo.getCommit(target);
 +        } else if (reference.isBranch() && reference.shorthand() === self.options.masterBranchName) {
 +          commitPromise = repo.getBranchCommit(name);
 +        } else {
 +          return;
 +        }
 +        self.log('info', 'Gotten a promise for commit for ', name);
 +        self.log('info', 'Adding a promise to get commit tree for ', name);
 +        treesDataPromises.push(buildTreePromise(commitPromise, name));
 +      });
 +
 +      self.log('info', 'Resolving commit tree promises');
 +      return Promise.all(treesDataPromises);
```

yomexzo: is using `var self = this` a bad pattern in the context described with the promises? or `.bind()` provides a benefit?

scripts/util/doc-builder.js
yomexzo: Interestingly, the article suggests this patter and i think it made sense when @minasmart suggested it. It's **Advanced mistake #4: okay, what if I want the result of two promises?**

yomexzo: @m-ux, i learnt about `Promise.resolve().then()` after reading the article. I had the other approach of creating new promises earlier. So we would still have absolutely sane code converting the functions to return promises. I'd get right on that.

Thanks.

scripts/util/doc-builder.js
minasmart: If you turned all of your nesting into functions, it'd become significantly cleaner:

```javascript
function filterNonTags(reference) {
  return (
    reference.isTag() ||
    (reference.isBranch() && reference.shorthand() === this.options.masterBranchName)
  );
}

function treeFromRef(reference) {
    var target = reference.targetPeel() || reference.target();
    var name = reference.shorthand();
    var commitPromise;

    if (reference.isTag()) {
      commitPromise = repo.getCommit(target);
    } else {
      commitPromise = repo.getBranchCommit(name);
    }

    this.log('info', 'Gotten a promise for commit for ', name);
    this.log('info', 'Adding a promise to get commit tree for ', name);

    return buildTreePromise(commitPromise, name);
}

function extractTags(references) {

    this.log('info', 'Filtering them to include only tags and the ', this.options.masterBranchName, ' branch.');
    var treesDataPromises = references.filter(filterNonTags.bind(this))
                                      .map(treeFromRef.bind(this));

    this.log('info', 'Resolving commit tree promises');
    return Promise.all(treesDataPromises);
}

function getReferencesForRepo(repo) {
  this.log('info', 'Getting available references within the repo');

  return repo.getReferences(NodeGit.Reference.TYPE.OID).then(extractTags.bind(this));
}

NodeGit.Repository.open('.').then(getReferencesForRepo.bind(this));
```

At that point, there's only one `this` to track, and every set of functionality can be pretty clearly represented as its own function.

m-ux: +1

scripts/util/doc-builder.js
m-ux: np - if you're using a library like (Bluebird)[http://bluebirdjs.com/docs/getting-started.html] you can also use the function `Bluebird.promisify()` to automatically convert a callback based library into a promise based one.  

Not sure if this helps in this specific situation, but both this & `promise.resolve()` are useful utilities in a lot of situations.

Only thing to keep in mind about `promise.resolve()` is that it doesn't handle the `reject` scenario, but in a lot of cases you don't need anything else.

scripts/util/doc-builder.js
yomexzo: @minasmart @m-ux, so based on [this](https://github.com/Shopify/javascript#2.5), i used arrow functions instead which eliminated the need to save `this` in `self`. What do you think?

scripts/util/doc-builder.js
yomexzo: thanks @m-ux, all functions have now been updated to return promises

yomexzo:
fixes #122

scripts/build
```diff
@@ -6,7 +6,6 @@ require('./util/work-from-root')();
 
 var fsExtra = require('fs-extra');
 var buildProject = require('./util/build-project');
-
```
minasmart: Why remove this line?

scripts/build-doc
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+/* globals require process */
```
minasmart: use `/* eslint-env node */` (I just figured this out) 😄

scripts/build-doc
```diff
@@ -0,0 +1,18 @@
...
+var docBuilder = require('./util/doc-builder')({
+  docsBranchName: 'gh-pages',
+  rmSrcDir: true,
+  commitDocs: true
+});
+
+console.log('Building ....');
+docBuilder.build().then(function () {
+  console.log('\nDone building');
+}).catch(function(error) {
```
minasmart: space before params. consistency with L14.

scripts/build
yomexzo: i didn't realize they should be separated. should they?

minasmart: The first set are dependencies with `require`. The second bit is vars. I think they make sense to keep separated? Thoughts?

yomexzo: well, that makes sense

scripts/build-doc
yomexzo: 😄  i saw that in your `graphql` branch. replacing it now.
i doubt the file is being "auto" linted tho. no complains when that line is removed.

minasmart: It's not. None of these are. I (or someone) needs to put together an eslint config and routine for these files.

yomexzo:
@Shopify/javascript, kindly 👀

scripts/build-doc
```diff
@@ -0,0 +1,18 @@
...
+  docsBranchName: 'gh-pages',
+  rmSrcDir: true,
+  commitDocs: true
+});
+
+console.log('Building ....');
+docBuilder.build().then(() => {
+  console.log('\nDone building');
+}).catch(error => {
+  console.error('error', error);
```
dan-menard: Does `'error'` also need the `\n`?

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+      var regex = new RegExp('/' + this.options.docsBranchName + '$'); //ensure reference name ends with the provided branch name
+
+      for (key in references) {
+        if (references[key].name().match(regex)) {
+          docsBranchReference = references[key];
+          if (!references[key].isRemote()) {
+            break;
+          }
+        }
+      };
```
dan-menard: This loop is a little tricky to follow, because it's doing two things (making sure at least one reference matches the regex, checking if a reference is remote). It might be a bit cleaner to split this up into two pieces:

```javascript
docsBranchReferences = references.filter(function(key) {
  return references[key].name().match(regex);
});

if (!docsBranchReferences.length) {
  throw new Error(this.options.docsBranchName + ' could not be found');
}

remoteDocsBranchReferences = docsBranchReferences.filter(function(reference) {
  return reference.isRemote();
});
docsBranchReference = remoteDocsBranchReferences[0]

if (docsBranchReference) {
  // lines 50–57
}
```

Looking at that now, I'm not sure it's any easier to follow, but I like suggesting functional alternatives when I see 'em, even if it's just an interesting exercise for me to write out :)

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+          console.info('info:', 'Staging \'', this.options.versionsDirName, '\' to \'', this.options.docsBranchName, '\'');
+          return index.addAll(`${this.options.versionsDirName}`, NodeGit.Index.ADD_OPTION.ADD_FORCE);
+        }).then(() => {
+          return repo.getReference(this.options.docsBranchName);
+        }).then(reference => {
+          var docsBranchResolvedName = reference.name();
+          console.info('info:', 'Blew up', this.options.docsBranchName, 'into', docsBranchResolvedName);
+          console.info('info:', 'Getting the staging area tree id');
+          return index.writeTree().then(treeId => {
+            var dateObject = new Date();
```
dan-menard: eep! Are you sure you want to do the date stuff yourself, rather than use a sane lib like momentJS?

scripts/util/doc-builder.js
yomexzo: this actually looks very much better.
i also felt that block could be written much better ... good catch.

scripts/util/doc-builder.js
yomexzo: i could do that, but i'm not sure if it's worth including an entire library.

lemonmade: Agreed, not worth bringing in a library for this.

scripts/build
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-/* globals require process */
+/* eslint-env node */
```
lemonmade: This project isn't using our ESLint config yet, right?

scripts/build-doc
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+/* eslint-env node */
+
+require('./util/work-from-root')();
+
+var docBuilder = require('./util/doc-builder')({
```
lemonmade: What version of node is your project set to use locally? Seems like you should be able to use things like `const` and `class` throughout these scripts.

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env node
+
+var path = require('path');
+var fs = require('fs');
+var yuidoc = require('./yuidoc');
+var NodeGit = require('nodegit');
+var util = require('util');
+var fsExtra = require('fs-extra');
+
+function DocBuilder(options) {
```
lemonmade: Look like it could be a class, if your version of node supports it.

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+  this.options = util._extend({
```
lemonmade: `_extend` is part of this thing's public API? Can we use `Object.assign` instead?

scripts/build
minasmart: It's not. We have a  branch underway to normalize the build system, then we'll start implementing that.

scripts/build-doc
minasmart: we support 0.12, but we're considering bumping to the LTS (4.4.7). When we do that, we'll adjust.

lemonmade: This is just for the project itself though, right? I'd probably set a version of node you expect to be run locally and use that to give you access to some of the more modern features.

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+    this.options.themeDir = path.join(this.options.docsDirName, 'yuidoc', 'theme');
+  }
+
+  this.options.apiDirPath = path.join(this.options.docsDirName, this.options.versionsDirName);
+}
+
+DocBuilder.prototype.ensureDocsBranchExists = function () {
+  console.info('info:', 'Ensuring \'' + this.options.docsBranchName + '\' branch exists');
+  return NodeGit.Repository.open('.').then(repo => {
+    return repo.getReferences(NodeGit.Reference.TYPE.OID).then(references => {
```
lemonmade: This comment goes for a few things in this file: with promises, you almost never want to do a `then` inside of another `then`. Instead, `return` this promise, and put the `then` as a chained `then` from the one you are in here. It flattens things out nicely.

scripts/build-doc
minasmart: It is, but it was originally set up for 0.12 support for the sake of contributors.

We're considering bumping that to the LTS, but this isn't the place to make that change.

scripts/util/doc-builder.js
lemonmade: IMO it's fine to do this as a single filter operation, and you can use `find` if you only want one:

```js
docsBranch = Object
  .keys(references)
  .map((key) => references[key])
  .find((branch) => branch.name().match(regex) && branch.isRemote());

if (docBranch == null) {
  throw new Error(...)
}
```

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+    fsExtra.removeSync(this.options.docsDirName);
+  }
+
+  fs.mkdirSync(this.options.docsDirName);
+
+  console.info('\ninfo:', 'Checking out', this.options.docsBranchName, 'branch');
+  return NodeGit.Repository.open('.').then(repo => {
+
+    console.info('info:', 'Getting', this.options.docsBranchName, '\'s latest commit');
+    return repo.getBranchCommit(this.options.docsBranchName).then(commit => {
```
lemonmade: Same thing about the promise pyramid of doom.

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,236 @@
...
+      references.forEach(reference => {
+        var target = reference.targetPeel() || reference.target();
+        var name = reference.shorthand();
+        var isMasterBranch = reference.isBranch() && name === this.options.masterBranchName;
+
+        if (!isMasterBranch && !reference.isTag()) {
+          return;
+        }
+
+        promise = promise.then(() => {
```
lemonmade: Does your work on the references need to be done sequentially, or can it be done in parallel?

scripts/util/yuidoc.js
```diff
@@ -0,0 +1,53 @@
...
+  options.nocode = true;
+  options.quiet = true;
+  options.themedir = path.dirname(require.resolve(path.join('yuidoc-lucid-theme', 'theme.json')));
+  options.helpers = [path.join(options.themedir, 'helpers', 'helpers.js')];
+
+module.exports = {
+  generate: function (_options) {
+    options = util._extend(options, _options);
+    var starttime = (new Date()).getTime();
+    var promise = Promise.resolve();
```
lemonmade: If these can be done in parallel, you can do `Promise.all` instead of chaining the promises in this way.

scripts/util/doc-builder.js
yomexzo: So, I originally did that. I had to have "global" variables that was either used / mutated in the different `thens`.

This approach prevents such situation and you can clearly see which parts of the code depend on which variables. @minasmart suggested this.

scripts/util/doc-builder.js
yomexzo: i've had times in which i "fry" my clone of the repo because of simultaneous changes to the repo. I'd like to think the "reads" can occur simultaneously, but i seriously doubt checking out asynchronously would be a good idea.

i'd explore and see how this can be better written.

scripts/util/doc-builder.js
minasmart: I suggested encapsulating your logic into more functions instead of using the giant tree, and creating scopes instead of globals when you had to persist state. Which means a combination of the two.

minasmart: And I'm still of the opinion if you need to create vars in the modules global scope, and assign them somewhere, it's going to be very hard to follow, or figure out what is assigned where.

scripts/util/doc-builder.js
minasmart: Yeah, libgit doesn't support parallel operations too well, because it relies on the state of the inner `.git` directory. When you have parallel operations, you very quickly get into a broken state.

scripts/util/doc-builder.js
lemonmade: Definitely dependent on the situation, I agree that having a global that gets updated is hard to follow. The other option is to return objects/ arrays from then that have all the state that is needed for subsequent actions, it's up to you whether you feel that makes it easier or harder to read.

scripts/util/doc-builder.js
lemonmade: 👍, I wish constructing a chained promise like this wasn't so gross but looks like it's necessary in this case.

scripts/util/yuidoc.js
yomexzo: done!

minasmart: This var isn't used...?

scripts/util/doc-builder.js
mikkoh: Hey @lemonmade for some reason I can't view the original commit. But I'm guessing the above code was something like:

```
function DocBuilder(options) {
  var objectThatBuildsDocs = {};

  return objectThatBuildsDocs;
}
```

Should there be a preference for classes at Shopify over functions that return objects? I've always liked doing the above since you can instantiate either by calling as a function or using a the `new` operator.

lemonmade: Yeah, I really like the factory approach, and have no problem with that in general (although using an uppercase name for the factory makes it a little more confusing, I'd assume it's going to be a constructor function in that case). Much of the JS/ CoffeeScript in the rest of Shopify is much more class-oriented, though, which largely comes from it being done by people who were more comfortable writing Ruby.

mikkoh:
A majority of the code in this PR is around creating + updating `gh-pages` did you explore using:
https://www.npmjs.com/package/gh-pages

scripts/util/doc-builder.js
```diff
@@ -0,0 +1,240 @@
...
+              console.info('info:', 'Done commiting API docs');
+            });
+          });
+        });
+      });
+    });
+  }
+
+  build() {
+    return this.ensureDocsBranchExists().then(() => {
```
mikkoh: @yomexzo I know we sat down to discuss this in person but I'm adding this here as a note. The following could be done to the `build` function and then the other functions could be cleaned up in a similar method.

1. You could get and save the `repo` to the class scope since all subsequent functions request the repo everytime. This would eliminate `NodeGit.Repository.open('.')` calls down the line. Also it'll make it so that the promise trees remain at one depth.
2. The promise tree in the build function could be made "flat"/or at one depth (example below)
3. Some of the function names could be made more descriptive to allow for the promise chain to be more semantic. So you should be able to literally read the code as "getTheRepo __THEN__ ensureDocsBranchExists __THEN__ checkoutDocsBranch..." (example below)

Example:
```javascript
build() {
  NodeGit.Repository.open('.').then((repo) => {
    this.repo = repo;
  })
  .then(this.ensureDocsBranchExists)
  .then(this.checkoutDocsBranch)
  .then(this.getLatestSrcPaths)
  .then(this.generateAPIDocsIntoGeneratedPaths)
  .then(() => {
    if (this.options.rmSrcDir) {
      console.info('\ninfo:', 'Removing source directories');
      paths.forEach(item => {
        var srcPath = path.join(item, this.options.srcDirName);
        console.info('info:', 'Removing', srcPath);
        fsExtra.removeSync(srcPath);
      });
    }

    if (this.options.commitDocs) {
      return this.commitAPIDocs();
    }
  });
}
```

Also maybe the last anonymous function which gets called could be named just so it's all readable.

yomexzo:
Closing this PR because i'd be doing some major rewrite including dropping committing to `gh-pages` since Github pages can now be served from the `docs` folder

PR #150, opened 2016-06-29
Added custom domain and deprecating myShopifyDomain in config

tanema:
Changed client config to accept a `domain` config as well as deprecated `myShopifyDomain` to allow for completely custom domains. I added a deprecated transforms flow so that myShopifyDomain would be transformed to a domain config and issue a warning. This allows requiredProperties to act as it did before the change.

Added a logger that tags logs with `[JS-BUY-SDK]` so that when we issue deprecation warnings, it is clear where it comes from. It is pretty basic right now and is not much more than a wrapper around `console`. 

**cc** @tessalt @minasmart

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -15,7 +15,7 @@ const Config = CoreObject.extend({
    * @param {String} attrs.appId The app whose listings the client will be
    * using. If you are just modifying a buy button, the buy-button's app id is
    * 6. Otherwise, obtain the app id of the app you're modifying or extending.
-   * @param {String} attrs.myShopifyDomain You shop's `myshopify.com` domain.
+   * @param {String} attrs.domain Your shop's full domain.
    */
   constructor(attrs) {
     Object.keys(this.deprecatedProperties).forEach(key => {
@@ -105,6 +105,16 @@ const Config = CoreObject.extend({
    * @public
    */
   domain: ''
+
+  /**
+   * The subdomain of myshopify.io that all the api requests will go to
+   * @attribute myShopifyDomain
+   * @default ''
+   * @type String
+   * @public
+   * @deprecated Use `config.domain` instead.
+   */
+  myShopifyDomain: ''
 });
 
 export default Config;
--- a/src/config.js
+++ b/src/config.js
@@ -17,6 +17,14 @@ const Config = CoreObject.extend({
    * @param {String} attrs.myShopifyDomain You shop's `myshopify.com` domain.
    */
   constructor(attrs) {
+    Object.keys(this.deprecatedProperties).forEach(key => {
+      const transformName = this.deprecatedProperties[key];
+      const transform = this[transformName];
+
+      if (attrs.hasOwnProperty(key)) {
+        transform(attrs[key], attrs);
+      }
+    });
     this.requiredProperties.forEach(key => {
       if (!attrs.hasOwnProperty(key)) {
         throw new Error(`new Config() requires the option '${key}'`);
@@ -27,8 +35,37 @@ const Config = CoreObject.extend({
   },
 
   /**
-   * The apiKey for authenticating against shopify. This is your api client's
-   * public api token. Not the shared secret. Set during initialation.
+   * An object with keys for deprecated properties and values as functions that
+   * will transform the value into a usable value.
+   * @attribute deprecatedProperties
+   * @default { myShopifyDomain: this.transformMyShopifyDomain }
+   * @type Object
+   * @private
+   */
+  deprecatedProperties: {
+    myShopifyDomain: 'transformMyShopifyDomain'
+  },
+
+  /**
+   * Transform the myShopifyDomain config to a domain config.
+   * @method transformMyShopifyDomain
+   * @static
+   * @private
+   * @param {String} subdomain The original subdomain on myshopify.com
+   * @param {Object} attrs. The config attributes to be transformed to a
+   * non-deprecated state.
+   * @return {Object} the transformed config attributes.
+   */
+  transformMyShopifyDomain(subdomain, attrs) {
+    console.warn('Config - ',
+       `myShopifyDomain is deprecated,
+       please use domain and provide the whole myshopify.com domain.`);
+    attrs.domain = `${subdomain}.myshopify.com`;
+    delete attrs.myShopifyDomain;
+  },
+
+  /**
+   * Properties that must be set on initializations
    * @attribute requiredProperties
    * @default ['apiKey', 'appId', 'myShopifyDomain']
    * @type Array
@@ -37,7 +74,7 @@ const Config = CoreObject.extend({
   requiredProperties: [
     'apiKey',
     'appId',
-    'myShopifyDomain'
+    'domain'
   ],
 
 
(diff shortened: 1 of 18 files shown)


minasmart:
I know I'm being the worst, but can the logger component be a separate PR? Even if you just break your  commits apart?

tanema:
I can do that yes, what is the best way to break it apart? It was a single commit. Also how do I manage the fact that the domain work uses the logger for warning if they are separate pull requests? Do I make 3 pull requests? I have so many questions!

tanema:
Okay breaking a commit apart isn't that bad. I broke it apart. since the logger changes are made on the config changes though I am not sure how to make it into 2 pull requests.

src/config.js
```diff
@@ -17,6 +18,14 @@ const Config = CoreObject.extend({
    * @param {String} attrs.myShopifyDomain You shop's `myshopify.com` domain.
    */
   constructor(attrs) {
+    Object.keys(this.deprecatedProperties).forEach(key => {
+      const transformName = this.deprecatedProperties[key];
```
tessalt: could move this stuff within the conditional so you're only doing it if there is a deprecated property

tanema: Good call, **done**

src/config.js
```diff
@@ -27,8 +36,37 @@ const Config = CoreObject.extend({
...
+   * @private
+   * @param {String} subdomain The original subdomain on myshopify.com
+   * @param {Object} attrs. The config attributes to be transformed to a
+   * non-deprecated state.
+   * @return {Object} the transformed config attributes.
+   */
+  transformMyShopifyDomain(subdomain, attrs) {
+    logger.warn('Config - ',
+       `myShopifyDomain is deprecated,
+       please use domain and provide the whole myshopify.com domain.`);
```
minasmart: `provide the full shop domain`. The point of this is that the merchant may wish to use their own CNAME, and also, with products like sello: we're introducing shops that are subdomains of other top level shopify domains.

src/config.js
```diff
@@ -27,8 +36,37 @@ const Config = CoreObject.extend({
   },
 
   /**
-   * The apiKey for authenticating against shopify. This is your api client's
-   * public api token. Not the shared secret. Set during initialation.
+   * An object with keys for deprecated properties and values as functions that
+   * will transform the value into a usable value.
```
minasmart: I would mention the expected signature of any given transform is `value, configHash`.

src/logger.js
```diff
@@ -0,0 +1,45 @@
...
+   * Also allows us to disable output in production.
+   * @class Logger
+   * @constructor
+   */
+  constructor() {
+  },
+
+  /* eslint-disable no-console */
+  log() {
+    console.log(...this.tagLog(...arguments));
```
minasmart: This may be one level too deep abstraction, but you could have a high order function that returns these:

```javascript

wrapConsole(logCommand) {
  return function() {
    console[logCommand](...this.tagLog(...arguments));
  };
}


// Then in your class
CoreObject.extend({
  log: wrapConsole('log'),
  debug: wrapConsole('debug')
})
```

Also, a thing I'm opinionated about (so it may be the wrong way to go): I don't like presenting the default console.log. I feel like if we're outputting something, we should be opinionated about what that something is. It's one of debug info, informative, warning, or errors. `console.log` is very ambiguous on intent.

tanema: I like that wrap method. It definitely keeps things DRY, it's even similar to how I tested the methods. As for `console.log` I can take it out. It doesn't really matter and also you are quite correct. We should be precise about logging intent if we are outputting to the console in an environment that isn't our own.

src/shopify.js
```diff
@@ -41,8 +42,9 @@ const Shopify = {
    * @param {String} configAttrs.appId The app whose listings the client will be
    * using. If you are just modifying a buy button, the buy-button's app id is
    * 6. Otherwise, obtain the app id of the app you're modifying or extending.
-   * @param {String} configAttrs.myShopifyDomain You shop's `myshopify.com`
+   * @param {String} configAttrs.myShopifyDomain [DEPRACATED] You shop's `myshopify.com`
```
minasmart: There's actually an API for `[DEPRECATED]`

image

From [here](http://yui.github.io/yuidoc/syntax/)

src/config.js
tanema: Good call I will make those changes.

tests/unit/api/config-test.js
```diff
@@ -32,3 +32,52 @@ test('it doesn\'t throw when all required params are specified', function (asser
...
+  const config = new Config({
+    myShopifyDomain: 'krundle',
+    apiKey: 123,
+    appId: 6
+  });
+
+  assert.equal(config.domain, 'krundle.myshopify.com', 'domain should be myshopify.com domain');
+});
+
+test('it output a deprecation warning when using myShopifyDomain', function (assert) {
```
minasmart: `it should`

tests/unit/api/config-test.js
```diff
@@ -32,3 +32,52 @@ test('it doesn\'t throw when all required params are specified', function (asser
...
+  assert.expect(1);
+
+  const config = new Config({
+    domain: 'krundle.com',
+    apiKey: 123,
+    appId: 6
+  });
+
+  assert.equal(config.domain, 'krundle.com', 'domain should be krundle.com domain');
+});
```
minasmart: This may be asking a lot, but do you think it's worthwhile testing that the deprecation pattern works? Like, take the config class, add a generic deprecated property, and add a generic transform, then verify that the deprecation loop thingy works? This may be hard to test, but maybe it's a good thing to do?

minasmart: What I'm getting at is that you've introduced a new design pattern, and we may be able to abstract it eventually and use it on other classes (like an extension to `CoreObject` or something). So it may be useful to expose that pattern, so if we do extract it, we have some loose assurances.

tests/unit/api/logger-test.js
```diff
@@ -0,0 +1,27 @@
...
+      assert.equal(arguments[0], '[JS-BUY-SDK]: ', `console.${method} should be tagged`);
+      assert.equal(arguments[1], testOutput, `console.${method} should be wrapped`);
+    };
+
+    logger[method](testOutput);
+    console[method] = oldLog;
+    /* eslint-enable no-console */
+  });
+});
+
```
minasmart: 🔥 empty line

minasmart:
I think this is really good, nitpicks aside. It's also one of the last things we need to release `v0.2.0` as per #51 

Address nitpicks, squash fixups into relevant commits, then :shipit:

tests/unit/api/config-test.js
tanema: I am always open to more testing, it always steadies my nerves so I have no problem doing more. As you say it would be valuable if this pattern is reusable.

src/config.js
```diff
@@ -14,9 +15,17 @@ const Config = CoreObject.extend({
    * @param {String} attrs.appId The app whose listings the client will be
    * using. If you are just modifying a buy button, the buy-button's app id is
    * 6. Otherwise, obtain the app id of the app you're modifying or extending.
-   * @param {String} attrs.myShopifyDomain You shop's `myshopify.com` domain.
+   * @param {String} attrs.domain Your shop's full domain.
```
richgilbank: It would be good to add an example here, so people don't associate it with the domains they add to their shop (if they're using Online Store). 
```diff
-   * @param {String} attrs.domain Your shop's full domain.
+   * @param {String} attrs.domain Your shop's full domain (i.e. `craftedgoods.myshopify.com`)
```

tanema:
**BUMP** for finishing reviews.

yomexzo:
good work @tanema 😸

PR #152, opened 2016-07-04
Prepended layout information to the yuidoc theme

yomexzo:
#149 depends on this PR


mikkoh:
@yomexzo since the theme is a separate module now do we still need this PR?

yomexzo:
No we don't. 🔥

PR #156, opened 2016-07-07
BugFix - setGuidFor currently uses numberCache instead of stringCache

yomexzo:
BugFix .. setGuidFor currently uses numberCache instead of stringCache when sent item is an object

### DIFF
--- a/src/metal/set-guid-for.js
+++ b/src/metal/set-guid-for.js
@@ -50,7 +50,7 @@ function setGuidFor(obj) {
         id = stringCache[obj];
 
         if (!id) {
-          id = numberCache[obj] = `st${uuid()}`;
+          id = stringCache[obj] = `st${uuid()}`;
         }
 
         break;


yomexzo:
cc @harismahmood89 @tessalt

yomexzo:
kindly 👀 @minasmart @tanema 
thanks

tanema:
Sure looks good but next time can you explain more of the problem, perhaps what `numberCache` was and why it should use `stringCache` now.

minasmart:
@tanema They exist because you can't stick a key to a literal. So it maintains hashes keyed by the literals for each type. The issue is that if you have `cache[1]` for the literal int `1`, and then `cache["1"]` for the literal string `"1"`, your hash will use the same ID for both the string and the int.

image

yomexzo:
what she said 😁 
but yes. i'd do that next time @tanema 
Thanks!

Guess that was your go @minasmart 😄

PR #158, opened 2016-07-08
Adding protection to the log wrapping.

tanema:
In IE9, apply cannot be called on console.log or any console method so I have wrapped the console methods in a function prototype from suggestion by @minasmart. I also added a try catch to catch if the method is undefined. It will fall back to console.log if the method being called is undefined.

cc @minasmart @tessalt

### DIFF
--- a/ci/deps
+++ b/ci/deps
@@ -35,4 +35,8 @@ fi
 
 ln -s $versioned_node_modules_path node_modules
 
+# See: https://github.com/npm/npm/issues/11283
+npm set progress=false
+
+# Install deps
 npm install
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,21 +1,21 @@
 import CoreObject from './metal/core-object';
 
 function wrapConsole(logCommand) {
-  let logMethod;
-
-  /* eslint-disable no-console */
-  try {
-    logMethod = Function.prototype.bind.call(console[logCommand], console);
-  } catch (e) {
-    logMethod = Function.prototype.bind.call(console.log, console);
-  }
-  /* eslint-enable no-console */
+  const logMethod = function () {
+    /* eslint-disable no-console */
+    if (console[logCommand]) {
+      console[logCommand](...arguments);
+    } else {
+      console.log(...arguments);
+    }
+    /* eslint-enable no-console */
+  };
 
   return function () {
-    const args = Array.prototype.slice.call(arguments);
+    const args = [...arguments];
 
     args.unshift('[JS-BUY-SDK]: ');
-    logMethod.apply(console, args);
+    logMethod(...args);
   };
 }
 
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,13 +1,21 @@
 import CoreObject from './metal/core-object';
 
 function wrapConsole(logCommand) {
+  let logMethod;
+
+  /* eslint-disable no-console */
+  try {
+    logMethod = Function.prototype.bind.call(console[logCommand], console);
+  } catch (e) {
+    logMethod = Function.prototype.bind.call(console.log, console);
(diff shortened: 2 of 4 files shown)


src/logger.js
```diff
@@ -6,7 +6,11 @@ function wrapConsole(logCommand) {
 
     args.unshift('[JS-BUY-SDK]: ');
     /* eslint-disable no-console */
-    console[logCommand](...args);
+    try {
+      Function.prototype.bind.call(console[logCommand], console).apply(console, args);
+    } catch (e) {
+      Function.prototype.bind.call(console.log, console).apply(console, args);
+    }
```
minasmart: I would do this once. Just define `log` at the top of the file, then call it in here. Doing a try/catch on every call is probably unnecessary.

tanema: good call

tests/unit/api/logger-test.js
```diff
@@ -1,47 +1,46 @@
...
-['debug', 'info', 'warn', 'error'].forEach(method => {
-  test(`it should wrap ${method}`, function (assert) {
-    assert.expect(2);
-
-    /* eslint-disable no-console */
-    const oldLog = console[method];
+test('it should wrap a valid console method', function (assert) {
+  assert.expect(2);
+  /* eslint-disable no-console */
+  const oldLog = console.log;
```
minasmart: Shouldn't the stubbing and unstubbing be run in a module's setup/teardown?

tanema: yeah I can do that it would clean it up a bit.

tanema:
Okay @tessalt give me a hallelujah and I will give you IE9 support 🌴

tessalt:
💯

tanema:
I cannot get circle to run succesfully on the PR I keep getting the error `Warning: ignoring input sourcemap for route-recognizer.js because ENOENT, no such file or directory '/home/ubuntu/js-buy-sdk/tmp/source_map_concat-input_base_path-ZA4H4wnO.tmp/0/route-recognizer.js.map'` and I didnt change anything in how things build or run tests.

minasmart:
That's just a missing source so warning. It happens in dev too. The error is likely past that. Have you done a rebuild without cache?

—
Mina Smart

minasmart:
OK. Got the test system fix into master, and rebased. The issue is with `function.bind` in phantom. I don't know that there's a fix for it at the moment, just an [epic issue](https://github.com/ariya/phantomjs/issues/10522) discussing it. I'm going to refactor without bind, and see how it goes.

mikkoh:
@tanema @minasmart was the original issue that something like `console.warn` didn't exist in ie9 or did it have drastically different behaviour where it actually threw an error? Trying to understand what the original issue without having to bring in `bind` shims just to make  Phantom work.

minasmart:
Hey! I have this branch almost fixed. It was totally ie9 though. I think it's absolutely possible without bind. I can show you what I have IRL if you want.

mikkoh:
Nice that's what I was thinking that we prob don't need bind. Nice 👍

tanema:
Oh so that was the problem all along, if only there were clear error messages to let me know that and I could have fixed this ten years ago.

minasmart:
ok, @tanema and @mikkoh this is probably good to go.

Also, if anyone knows how to disable the npm 2.x spinner, please let me know.

ci/deps
```diff
@@ -35,4 +35,8 @@ fi
 
 ln -s $versioned_node_modules_path node_modules
 
+# See: https://github.com/npm/npm/issues/11283
+npm set progress=false
```
minasmart: Not really related to this PR, but useful.

PR #160, opened 2016-07-08
Fix/versioner and licenser to handle directories and include non-js files in output path

yomexzo:
👀 @tessalt @tanema @minasmart


build-lib/util/recursive-read-dir.js
```diff
@@ -0,0 +1,21 @@
...
+
+function recursiveReadDir (dir) {
+  let files = [];
+
+  fs.readdirSync(dir).forEach(fileName => {
+    const file = path.join(dir, fileName);
+    if (fs.statSync(file).isDirectory()) {
+      files = files.concat(recursiveReadDir(file));
+    } else {
+      files.push(file);
```
minasmart: use filter and map. `forEach` is only for side effects.

build-lib/util/licenser.js
```diff
@@ -21,22 +25,20 @@ Licenser.prototype.constructor = Licenser;
...
-    });
-
-    return fileAcc.concat(list);
-  }, []).filter(fileName => {
-    return fileName.match(/^.+\.js$/);
-  });
-
-  files.forEach(fileName => {
-    const inputBuffer = fs.readFileSync(path.join(fileName));
+  this.inputPaths.forEach(dir => {
```
minasmart: This should be `const fileNames = [].concat(...this.inputPaths.map(recursiveReadDir))` followed by `fileNames.forEach(// write file)`. No need to conflate building up your data sets with side effects.

build-lib/util/recursive-write-file.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+const fs = require('fs');
+const path = require('path');
+
+function ensureDirExists (fullPath) {
```
minasmart: same here with spacing around the arguments

build-lib/util/recursive-write-file.js
```diff
@@ -0,0 +1,26 @@
...
+      fs.mkdirSync(fullPath);
+    } else {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+module.exports = function (fullPath, content) {
```
minasmart: anonymous functions have the space. Don't change this.

build-lib/util/recursive-write-file.js
```diff
@@ -0,0 +1,26 @@
...
+  if(ensureDirExists(path.dirname(fullPath))) {
+    fs.writeFileSync(fullPath, content);
+    return true;
+  }
+
+  return false;
+};
```
minasmart: What is recursive about this? I feel like you should just have a `createDirectoryIfNoneExists` function that takes the path, then write files as normal.

build-lib/util/versioner.js
```diff
@@ -59,28 +61,23 @@ Versioner.prototype.replaceTemplateString = function (versionString, buffer) {
 Versioner.prototype.versionFiles = function () {
   const versionString = `v${this.constructor.VERSION}-${this.constructor.HEAD_COMMIT}`;
 
-  const files = this.inputPaths.reduce((fileAcc, dir) => {
-    const list = fs.readdirSync(dir).map(fileName => {
-      return path.join(dir, fileName);
+  this.inputPaths.forEach(dir => {
+    const fileNames = recursiveReadDir(dir);
```
minasmart: Same here regarding map and forEach.

build-lib/util/licenser.js
```diff
@@ -21,22 +25,20 @@ Licenser.prototype.constructor = Licenser;
...
+    const fileNames = recursiveReadDir(dir);
```
minasmart: new line after assignments and before returns

build-lib/util/recursive-read-dir.js
```diff
@@ -4,18 +4,19 @@ const fs = require('fs');
...
 
-  fs.readdirSync(dir).forEach(fileName => {
+  const filesArray = fs.readdirSync(dir).map(fileName => {
     const file = path.join(dir, fileName);
+
     if (fs.statSync(file).isDirectory()) {
-      files = files.concat(recursiveReadDir(file));
-    } else {
-      files.push(file);
+
```
minasmart: Doesn't need a line break when it's the only thing in the body of `{ ... }`

build-lib/util/recursive-read-dir.js
```diff
@@ -4,18 +4,19 @@ const fs = require('fs');
...
+      return files.concat(recursiveReadDir(file));
```
minasmart: `files` isn't defined in this scope.

build-lib/util/recursive-write-file.js
yomexzo: true!

build-lib/util/licenser.js
yomexzo: i just took a shot at rewriting and might have to scope the file writing while going through the inputs. The `writeFile` bit needs the base directory name from which it was created

yomexzo:
changes made ... 👀 @tessalt @minasmart 
thanks

build-lib/util/licenser.js
yomexzo: nvm

build-lib/util/licenser.js
```diff
@@ -25,7 +25,7 @@ Licenser.prototype.constructor = Licenser;
 
 Licenser.prototype.build = function () {
 
-  const fileDescriptions = [].concat(...this.inputPaths.map(dirname => {
```
minasmart: These files are loaded under babel-register, so use the spread operator.

yomexzo: i replaced it because the build was failing for the reason

yomexzo: i mean on CI. Wondered why., installed 4.2.2 locally and got same. I just checked and babel register is on for versions less than `v4.0.0`

minasmart: so maybe bump the babel-register version check?

yomexzo: yeah. done and pushed. kindly 👀

yomexzo: bumped to support at least 4.4.7, spread operator became available natively after that.

yomexzo:
hey @mikkoh, @tanema, @tessalt .. kindly 👀 whenever you have some time ..  It's a fairly some change

## Some context
This PR modifies the `versioner` and  the`licenser` broccoli plugins to support handling of deep files buried in sub-directories. They are responsible for adding version and license information, respectively, to the top of JS files and they are [called here](https://github.com/Shopify/js-buy-sdk/blob/master/build-lib/lib.js#L78).

The thing is broccoli trees (directories) can have sub-trees (directories within them). They currently safely ignore sub directories because they only pick files on the "root" directory. See [here](https://github.com/Shopify/js-buy-sdk/blob/master/build-lib/util/versioner.js#L63) and [here](https://github.com/Shopify/js-buy-sdk/blob/master/build-lib/util/licenser.js#L26).

mikkoh:
**tldr** Maybe this is not needed and should not be merged in.

To be able to recursively add headers (license + version) to multiple files is a neat feature. However the only real use case for this to add header information to the Node build.

Typically in the Node ecosystem licenses are not distributed in JS source but is defined in the `package.json` and external in a `LICENSE` file. The version of the application is defined in the `package.json` file and should correspond to `git tags` anyway.

So. Since this feature would only benefit the Node build and it's not standard to the node ecosystem I'm not sure we need it right now.

yomexzo:
Hmm ... Good points.

Another angle to see this from is that they are `broccoli` plugins. Without these, it means the plugins would safely ignore some files and would not pass them onto the "next plugin". One might argue it's the last phase of the build process. But you can never tell for someone contributing and is not aware of what's going on within the versioner and the licenser.

Either ways is fine by me.

mikkoh:
This pull request will be "fixed" but handled slightly differently in #197

mikkoh:
@yomexzo @minasmart I'm closing this since things have changed drastically

minasmart:
👍🏻

PR #161, opened 2016-07-08
Rewrote the promise shim to be usable in a node environment.

yomexzo:
👀 @tessalt @tanema @minasmart

### DIFF
--- a/shims/promise.js
+++ b/shims/promise.js
@@ -1,4 +1,21 @@
-const RSVP = window.RSVP;
+/* globals require */
+
+function isNodeLikeEnvironment() {
+  const windowAbsent = typeof window === 'undefined';
+  const requirePresent = typeof require === 'function';
+
+  return windowAbsent && requirePresent;
+}
+
+let RSVP;
+
+if (isNodeLikeEnvironment()) {
+  const localRequire = require;
+
+  RSVP = localRequire('rsvp');
+} else {
+  RSVP = window.RSVP;
+}
 const Promise = RSVP.Promise;
 
 export { RSVP, Promise };


shims/promise.js
```diff
@@ -1,4 +1,21 @@
...
+
+let RSVP;
+
+if (isNodeLikeEnvironment()) {
+  const localRequire = require;
+
+  RSVP = localRequire('rsvp');
+} else {
+  RSVP = window.RSVP;
+}
```
minasmart: So, weird thing about this: I'm pretty sure nothing is actually using this shim? Nothing imports promise or RSVP. What exactly needs it?

minasmart: Can we just remove the shim altogether?

yomexzo: hmm... They are used in tests [here](https://github.com/Shopify/js-buy-sdk/blob/master/tests/unit/adapters/listings-adapter-test.js#L4) and [here](https://github.com/Shopify/js-buy-sdk/blob/master/tests/unit/api/shop-client-test.js#L5)

minasmart: Shouldn't be. Just remove the import and use the global promise. Then this shim can go away 😄

yomexzo: I'm thinking, Won't that break in environments like IE10?

minasmart: no, because it's polyfilled still. This is just a shim so that you can `import`

yomexzo: i see it. thanks!

yomexzo:
Will be creating another pull request to delete the promise shim

PR #163, opened 2016-07-11
Removed promise & fetch shim from being added to build since they are now polyfilled

yomexzo:
👀 @richgilbank @minasmart @tessalt

### DIFF
--- a/build-lib/lib.js
+++ b/build-lib/lib.js
@@ -18,12 +18,7 @@ const Versioner = require('./util/versioner');
 function sourceTree(pathConfig, moduleType) {
   const lib = babelTranspiler(pathConfig.lib, babelConfig(pkg.name, moduleType));
 
-  const shims = babelTranspiler(
-    funnel(pathConfig.shims, { include: ['fetch.js', 'promise.js'] }),
-    babelConfig(null, moduleType)
-  );
-
-  return mergeTrees([lib, shims]);
+  return mergeTrees([lib]);
 }
 
 module.exports = function (pathConfig, env) {
--- a/build-lib/lib.js
+++ b/build-lib/lib.js
@@ -16,9 +16,7 @@ const Versioner = require('./util/versioner');
 
 
 function sourceTree(pathConfig, moduleType) {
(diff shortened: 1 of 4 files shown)


build-lib/lib.js
```diff
@@ -18,12 +18,7 @@ const Versioner = require('./util/versioner');
 function sourceTree(pathConfig, moduleType) {
   const lib = babelTranspiler(pathConfig.lib, babelConfig(pkg.name, moduleType));
 
-  const shims = babelTranspiler(
-    funnel(pathConfig.shims, { include: ['fetch.js', 'promise.js'] }),
-    babelConfig(null, moduleType)
-  );
-
-  return mergeTrees([lib, shims]);
```
minasmart: You've also removed the fetch polyfill. Was that intentional? Is it used anywhere?

yomexzo: yeah, `fetch.js` does not exist. I think you removed it when you moved it to being polyfilled way back.

minasmart: Maybe just adjust the PR title, so we know the full effects then.

minasmart: And remove `fetch.js` from the shims directory.

yomexzo: `fetch.js` does not exist in the shims directory. You removed it already.

yomexzo: i think you did rather. But it's not there. Editing the title now.

yomexzo:
The Promise shim is not needed any more since `Promise` is being polyfilled [here](https://github.com/Shopify/js-buy-sdk/blob/master/build-lib/polyfills.js#L17). Fetch shim is also being polyfilled.

build-lib/lib.js
```diff
@@ -18,12 +18,7 @@ const Versioner = require('./util/versioner');
...
+  return mergeTrees([lib]);
```
minasmart: You can just return lib here, since it's one tree.

yomexzo: true. done. thanks.

minasmart:
Can you rebase and get this in?

Issue #153, opened 2016-07-05
"_global2.default.require is not a function" error on Node - Introduced in version 0.1.7


yoadsn:
Hey guys,
I have upgraded to 0.1.7 yesterday and started experiencing the following error when the node application runs (I use Shopify Buy SDK both on the server and on the client):

```
d:\Development\Experiments\tmp_node_test\node_modules\shopify-buy\lib\isomorphic-fetch.js:16
  _global2.default.fetch = _global2.default.require('node-fetch');
                                            ^

TypeError: _global2.default.require is not a function
    at Object.<anonymous> (d:\Development\Experiments\tmp_node_test\node_modules\shopify-buy\lib\isomorphic-fetch.js:16:45)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (d:\Development\Experiments\tmp_node_test\node_modules\shopify-buy\lib\shopify.js:19:1)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
```

I have created a minimal possible scenario to make sure it is not something specific to my code:
1. empty folder - `npm init`
2. `npm install shopify-buy -- save`
3. created index.js with the following contents:
`var s = require("shopify-buy");`
4. Running `node index` produces the above error

Node version used: v5.6.0
Npm version used: 3.10.2

From my investigation, the `global.require` access used instead of globally using `require` (pre 0.1.7) is the culprit. For some reason, when running node as a REPL - `global.require` is defined. But when running it with a script - `global.require` is not defined.

So far, digging into node's repl and c++ code I could not find any reason for this difference. Also I am currently not able to test on node V4.2.2 which is the highest version your CI tests on.

@minasmart  Is there a reason you chose to not use the globally available "require" (after you confirmed it's a "node like environment") ?

minasmart:
Hey! 

Thanks for investigating this so thoroughly. The problem with using `require` on its own is only that babel chokes on it. It's module compilers get confused when they see require conditionally nested. I'll look into this right away and see if there's a better workaround. I'm also open to recommendations if you think you have some insight into what the root of the problem may be.

Thanks Again!

yoadsn:
[...]

PR #164, opened 2016-07-11
Updated isomorphic-fetch's use of global.require because it breaks in node env

yomexzo:

### DIFF
--- a/src/isomorphic-fetch.js
+++ b/src/isomorphic-fetch.js
@@ -6,8 +6,8 @@ import isNodeLikeEnvironment from './metal/is-node-like-environment';
 const fetch = global.fetch;
 
 if (!fetch && isNodeLikeEnvironment()) {
-  /* global.require is only available in the REPL
-   * and Babel fails when you use require() explicitly
+  /* this indirection is needed because babel throws errors when
+   * transpiling require('node-fetch') using `amd` plugin with babel6
    */
   const localRequire = require;
 
--- a/src/isomorphic-fetch.js
+++ b/src/isomorphic-fetch.js
@@ -6,7 +6,7 @@ import isNodeLikeEnvironment from './metal/is-node-like-environment';
 const fetch = global.fetch;
 
 if (!fetch && isNodeLikeEnvironment()) {
-  /* global.require is only available in the REPL 
+  /* global.require is only available in the REPL
    * and Babel fails when you use require() explicitly
    */
   const localRequire = require;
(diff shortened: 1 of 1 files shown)


src/isomorphic-fetch.js
```diff
@@ -1,9 +1,13 @@
+/* globals require */
+
 import global from './metal/global';
 import isNodeLikeEnvironment from './metal/is-node-like-environment';
 
 const fetch = global.fetch;
 
 if (!fetch && isNodeLikeEnvironment()) {
-  global.fetch = global.require('node-fetch');
+  const localRequire = require;
```
minasmart: Can you verify that this indirection is sufficient to pass `npm run build`? I would also annotate the reason for the indirection.

yomexzo: yeah. verified.
added comment too.

src/isomorphic-fetch.js
```diff
@@ -1,9 +1,16 @@
...
 import global from './metal/global';
 import isNodeLikeEnvironment from './metal/is-node-like-environment';
 
 const fetch = global.fetch;
 
 if (!fetch && isNodeLikeEnvironment()) {
-  global.fetch = global.require('node-fetch');
+  /* global.require is only available in the REPL 
+   * and Babel fails when you use require() explicitly
+   */
```
minasmart: More than babel failing, we can assign this specifically to the plugin `babel-plugin-transform-es2015-modules-amd`

yomexzo: hmm... i'm wondering, is it `amd` or `commonjs`? i thought the `commonjs` versions were loaded when the library is required/used in the nodejs environment.

minasmart: You can repeatably cause this bug under the amd processor. The commonjs one seems fine. There's also a long outstanding bug which I couldn't find documenting it only on the amd processor.

yoadsn: I can confirm that on my side.

yomexzo: @minasmart, i get what you mean now. the problem also happens when transpiling to `umd`. It's a problem with babel6. I found situations when it choked too when transpiling scripts with `decorators`. I have updated the comment now

yomexzo:
@m-ux, @minasmart .. kindly 👀 

this change is needed because `global.require` yields `undefined` when the library is being used in node.

i had to use an intermediate variable instead of just using `require()` because transpiling to `amd` or even `umd` with `babel6` fails when `require` is used explicitly with an error with message like `TypeError: src/isomorphic-fetch.js: Property right of AssignmentExpression expected node to be of a type ["Expression"] but instead got null`

fixes #153

minasmart:
Can this 🚢 ?

PR #165, opened 2016-07-11
Store now looks for localStorage in global instead of in window object

yomexzo:


minasmart:
👍

tessalt:
lgtm

PR #166, opened 2016-07-11
Made cart-model and test look for ga in global instead of window object

yomexzo:

This is needed for tests to run in node

### DIFF
--- a/tests/unit/models/cart-model-test.js
+++ b/tests/unit/models/cart-model-test.js
@@ -1,3 +1,5 @@
+/* globals global */
+
 import { module, test } from 'qunit';
 import CartModel from 'shopify-buy/models/cart-model';
 import BaseModel from 'shopify-buy/models/base-model';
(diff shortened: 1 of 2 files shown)


tests/unit/models/cart-model-test.js
```diff
@@ -312,8 +314,15 @@ test('it detects google analytics and appends the cross-domain linker param', fu
   assert.equal(model.checkoutUrl, `${baseUrl}/${variantId}:${quantity}?${query}`);
 
   const variant = singleProductFixture.product_listing.variants[1];
+  let globalScope;
```
minasmart: Don't duplicate this logic. Just include the global scope helper from `src/metal`

yomexzo: yeah, i thought about that, i wasn't sure if it was ideal to include that since it's not being tested so i was waiting for feedback. I guess i'd just include it  already.

Thanks

yomexzo: updated now @minasmart

minasmart:
👍

richgilbank:
:shipit:

PR #167, opened 2016-07-11
Made sure btoa is loaded in nodejs environment when testing listings adapter

yomexzo:
Because `btoa` is available in browsers, tests run smoothly in PhantomJS but breaks when attempt is made to run the `listings-adapter` test in Node.

This imports `isomorphic-btoa` which ensures `btoa` is available in both environments.

### DIFF
--- a/tests/unit/adapters/listings-adapter-test.js
+++ b/tests/unit/adapters/listings-adapter-test.js
@@ -2,6 +2,7 @@ import { module, test } from 'qunit';
 import ListingsAdapter from 'shopify-buy/adapters/listings-adapter';
 import version from 'shopify-buy/version';
 import Promise from 'promise';
+import 'shopify-buy/isomorphic-btoa';
 
 let adapter;
 


yomexzo:
👀 @tanema @tessalt @minasmart

tests/unit/adapters/listings-adapter-test.js
```diff
@@ -2,6 +2,7 @@ import { module, test } from 'qunit';
 import ListingsAdapter from 'shopify-buy/adapters/listings-adapter';
 import version from 'shopify-buy/version';
 import Promise from 'promise';
+import 'shopify-buy/isomorphic-btoa';
```
minasmart: Should this be a concern here? Or farther up the chain? I feel like we shouldn't force the test to be aware of the isomorphs. Should we maybe instead include this behaviour in whatever file in the node tests does the equivalent of [this code]() does? https://github.com/Shopify/js-buy-sdk/blob/master/tests/index.html#L18-L22

yomexzo: I think that is the better way to go. I already have some wiring done in my node-test WIP, makes sense to add this to that since `btoa` might be needed in the future anywhere in the tests.

`src` will be fine since that's already handled in `shopify.js`

yomexzo:
Closing this PR in favour of adding loading of `btoa` centrally before tests are ran.

PR #171, opened 2016-07-14
Simplify the build loop process.

minasmart:
This PR takes a lot of what was statically coded, and turns it into data. The
biggest invisible change is that now we're building all the things all the time,
instead of only building some things during dev, and all the things an build.

This also pulls the AMD loader out of the dev built artifact, and makes it something that must be explicitly loaded into the test environment.

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -1,24 +1,25 @@
 /* global require, module */
 
-if (process.version < 'v4.0.0') {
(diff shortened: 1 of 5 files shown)


minasmart:
Please have a look see @tessalt and @yomexzo

tessalt:
i mean it looks good to me

PR #174, opened 2016-07-18
Schema generation

minasmart:
#This generates a schema that's consumable by a query builder thingy. It allows for dynamic fields and relationships, and gives you early warnings about things changing on the server side.

The product schema thing looks like this now.

```javascript
{
  "name": "Product",
  "fields": [
    "createdAt",
    "handle",
    "id",
    "productType",
    "publishedAt",
    "tags",
    "title",
    "updatedAt",
    "vendor"
  ],
  "fieldsWithArgs": {},
  "relationships": {
    "collections": {
      "type": "CollectionConnection",
      "args": [
        "first",
        "after",
        "reverse"
      ],
      "schemaModule": "collection-connection"
    },
    "images": {
      "type": "Image",
      "args": [
        "first",
        "maxWidth",
        "maxHeight",
        "crop",
        "scale"
      ],
      "schemaModule": "image"
    },
    "options": {
      "type": "ProductOption",
      "args": [
        "first"
      ],
      "schemaModule": "product-option"
    },
    "variants": {
      "type": "ProductVariant",
      "args": [
        "first"
      ],
      "schemaModule": "product-variant"
    }
  },
  "fieldOf": [
    {
      "type": "QueryRoot",
      "schemaModule": "query-root"
    },
    {
      "type": "Collect",
      "schemaModule": "collect"
    },
    {
      "type": "ProductEdge",
      "schemaModule": "product-edge"
    }
  ]
}
```

__NOTE:__ This is only for experimentation within the `graphql` branch. Not being merged into master.


minasmart:
@tessalt @lreeves have a look?

build-lib/lib.js
```diff
@@ -23,7 +23,13 @@ function sourceTree(pathConfig, moduleType) {
     babelConfig(null, moduleType)
   );
 
-  return mergeTrees([lib, shims]);
+  const schema = babelTranspiler(
+    (new GraphSchema()),
+    babelConfig('graph', moduleType)
+  );
+
+
```
lreeves: So much space for activities!

graph/fetch-schema.js
```diff
@@ -0,0 +1,17 @@
+import fetch from 'node-fetch';
+import introspectionQuery from './introspection-query';
+
+const baseGraphUrl = 'https://graphql.myshopify.com/api/graph';
+
+export default function fetchSchema() {
+  return fetch(baseGraphUrl, {
+    method: 'POST',
+    headers: {
+      Authorization: 'Basic MzUxYzEyMjAxN2QwZjJhOTU3ZDMyYWU3MjhhZDc0OWM=',
```
tessalt: we should probably put this somewhere else

graph/generator.js
```diff
@@ -0,0 +1,166 @@
...
+    /* eslint-disable no-underscore-dangle */
+    return extractTypeData(schema.data.__schema.types).filter(rejectBuiltins);
+    /* eslint-enable no-underscore-dangle */
+  };
+
+  if (cachedSchema) {
+    return Promise.resolve(cachedSchema).then(processSchema).then(exportTypesToFiles);
+  }
+
+  return fetchSchema().then(cacheSchema).then(processSchema).then(exportTypesToFiles);
```
tessalt: `catch`?

graph/fetch-schema.js
minasmart: It's not a secret, so I'm not tooooo worried about it. Ideally I'd like to have a common config that devs can use where they can pull the schema from their own shop. Until then: recommendation for where this should go? Or OK here? (this branch is prototype-y)

graph/generator.js
minasmart: None of the errors here are recoverable, so a catch would just be better error reporting. Is that what you're after?

tessalt: yep. having been bitten by useless promise errors lately.

minasmart: Cool. I'll add in some simple reporting.

lreeves:
I'm probably not the right person to pour through the ES stuff but overall the GraphQL stuff looks great 👍

graph/fetch-schema.js
tessalt: ok for now

PR #181, opened 2016-07-27
Graph Query Helpers

minasmart:
This sort of makes the process of writing graph queries easier. The
syntax is more verbose than I'd like, so I'm open to suggestions on a
better approach.

## What does it provide?
- A guarantee that if you're code works, you're asking for graph API data
that exists.
- Don't know what fields exist on a resource? Don't know resource
relationships? Look at the schema module in the dev console and
introspect all the things, OR use helpers to inject all the things into
your requests.
- Paginated requests. You don't want to have to write your own
nodes/edges/pageInfo fetching code, so this wraps that up.

## Example Usage
```javascript
const query = q('shop', shop => {
  const shopFields = fields(shop);
  const products = relationship(shop, 'products', { first: 10 }, product => {
    return fields(products);
  });

  return join(shopFields, products);
});
```

Query will be a string like
```graphql
query {
  shop {
    name,
    language,
    country,
    etc.
    products (first:10) {
      pageInfo {
        hasNextPage,
        hasPreviousPage
      },
      edges {
        cursor,
        node {
          name,
          handle,
          createdAt,
          updateAt,
          etc.
        }
      }
    }
  }
}
```

### DIFF
--- a/src/graph-helpers/join.js
+++ b/src/graph-helpers/join.js
@@ -1,0 +1,3 @@
+export default function join(...fields) {
+  return fields.join(', ');
+}
--- a/src/graph-helpers/relationship.js
+++ b/src/graph-helpers/relationship.js
@@ -1,0 +1,39 @@
+import relationship from './raw-relationship';
+import { parseArgs } from './raw-relationship';
+import fields from './fields';
+import join from './join';
+
+function wrapCallback(callbackFunction) {
+  return function (schema) {
+    if (schema.name.match(/Connection$/)) {
+      const pageInfo = relationship(schema, 'pageInfo', pageInfoSchema => {
+        return fields(pageInfoSchema);
+      });
+      const edge = relationship(schema, 'edges', edgeSchema => {
+        const edgeFields = fields(edgeSchema);
+        const node = relationship(edgeSchema, 'node', nodeSchema => {
+          return callbackFunction(nodeSchema);
+        });
+
+        return join(edgeFields, node);
+      });
+
+      return join(pageInfo, edge);
+    }
+
+    return callbackFunction(...arguments);
+  };
+}
+
+export default function connectionAwareRelationship(/* schema, relationshipKey, requestArgsHash, bodyCallback */) {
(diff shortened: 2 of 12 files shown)


minasmart:
@tessalt @lreeves @yomexzo have a look?

src/graph-helpers/join.js
```diff
@@ -0,0 +1,3 @@
+export default function join(...fields) {
+  return fields.join(', ');
+}
```
minasmart: This is for joining queries. Not fields. I realized just now it looks weird right next to the fields helper. It should probably be used there.

src/graph-helpers/relationship.js
```diff
@@ -0,0 +1,39 @@
...
+      const pageInfo = relationship(schema, 'pageInfo', pageInfoSchema => {
+        return fields(pageInfoSchema);
+      });
+      const edge = relationship(schema, 'edges', edgeSchema => {
+        const edgeFields = fields(edgeSchema);
+        const node = relationship(edgeSchema, 'node', nodeSchema => {
+          return callbackFunction(nodeSchema);
+        });
+
+        return join(edgeFields, node);
```
yomexzo: why not `[edgeFields, node].join(', ')`?
saves the need of having to create a new module.

lreeves:
Very much out of my wheelhouse but LGTM 👍 💯 🌷

src/graph-helpers/relationship.js
minasmart: because at that point, I'm relying on the implementer to use the same join everywhere. I would need everyone working on this to guarantee that the join characters are `', '`, which is not something I want to have to do, or build a linting rule around. This normalizes the pattern so nobody needs to care about what characters to use when joining.

minasmart: Additionally, knowing what characters are supported for joints between fields involves reading the graph spec, which is another thing I don't want anyone to have to do.

yomexzo: ah okay. i get you

yomexzo:
:shipit:

PR #183, opened 2016-08-02
Fix `npm run test-examples` failing on v6.2.2 due to stream event data sent as `Buffer` instead of `String`

yomexzo:
A child process sends streams on data events as against a string and the complete output is only available on completion of the stream.

This PR fixes this in `scripts/test-examples`. `npm run test-examples` failed on my machine because `.match` (a string function) is not a valid function on a stream buffer object.

👀 @minasmart @timbeiko

### DIFF
--- a/scripts/test-examples
+++ b/scripts/test-examples
@@ -33,7 +33,9 @@ function bootSelenium() {
     var selenium = childProcess.exec('java -jar selenium-standalone.jar');
 
     // Selenium outputs everything on STDERR because it's hateful
-    selenium.stderr.on('data', function (data) {
+    selenium.stderr.on('data', function (rawData) {
+      var data = rawData.toString();
+
       if (data.toString().match(/Selenium Server is up and running/) && !resolved) {
         resolved = true;
         resolve(selenium);
--- a/scripts/test-examples
+++ b/scripts/test-examples
@@ -61,11 +61,16 @@ function bootBroccoli() {
       examplesEnv.EXAMPLES = 1;
 
       var broccoli = childProcess.exec('node ' + broccoliBin + ' serve', { env: examplesEnv });
+      var broccoliOutput = "";
 
       broccoli.stdout.on('data', function (data) {
-        log('broccoli', data);
+        broccoliOutput += data;
+      });
 
-        if (data.match(/Serving on http/) && !resolved) {
+      broccoli.stdout.on('end', function () {
+        log('broccoli', broccoliOutput);
+
+        if (broccoliOutput.match(/Serving on http/) && !resolved) {
           resolved = true;
           resolve(broccoli);
         }
--- a/scripts/test-examples
+++ b/scripts/test-examples
@@ -62,8 +62,9 @@ function bootBroccoli() {
 
       var broccoli = childProcess.exec('node ' + broccoliBin + ' serve', { env: examplesEnv });
 
-      broccoli.stdout.on('data', function (data) {
-        data = data.toString();
+      broccoli.stdout.on('data', function (rawData) {
+        var data = rawData.toString();
+
         log('broccoli', data);
 
         if (data.match(/Serving on http/) && !resolved) {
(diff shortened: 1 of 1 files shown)


yomexzo:
👀 @mikkoh

scripts/test-examples
```diff
@@ -61,11 +61,16 @@ function bootBroccoli() {
       examplesEnv.EXAMPLES = 1;
 
       var broccoli = childProcess.exec('node ' + broccoliBin + ' serve', { env: examplesEnv });
+      var broccoliOutput = "";
 
       broccoli.stdout.on('data', function (data) {
-        log('broccoli', data);
```
minasmart: We shouldn't strip the logging.

minasmart: output doesn't necessarily end either. The approach of adding to the buffer is great, but I wouldn't assume we get an end event.

yomexzo: Well, the `end` event is certain. That's based on [this](https://nodejs.org/api/stream.html#stream_event_data) and [this](https://nodejs.org/api/stream.html#stream_event_end). I believe it makes sense to process it when we have the complete output.

In terms of where to put the logging, i'm not sure if it matters on either `data` or `end`

minasmart: `end` only occurs when there is no more data. Broccoli will continuously emit data. From the linked spec:

> The 'end' event is emitted when there is no more data to be consumed from the stream.

This only occurs if the process quits or closes STDOUT, which it doesn't.

yomexzo: haha .. i see what you mean 😄

scripts/test-examples
```diff
@@ -63,6 +63,7 @@ function bootBroccoli() {
       var broccoli = childProcess.exec('node ' + broccoliBin + ' serve', { env: examplesEnv });
 
       broccoli.stdout.on('data', function (data) {
+        data = data.toString();
```
minasmart: Don't reassign like this. Use a new var. Don't obfuscate where data comes from.

timbeiko:
@yomexzo I'm really not familiar with this part of the codebase. probably better to ping someone else to review. Sorry !

minasmart:
Is this change still necessary?

yomexzo:
oops @timbeiko ... i meant to ping @tanema.

@minasmart, it breaks in my environment - v6.2.2. May be you can try to replicate and see.

`npm run test-ci` or `node scripts/test-examples`

```
[broccoli] Serving on http://localhost:4200

/sites/shopify/js-buy-sdk/scripts/test-examples:68
        if (data.match(/Serving on http/) && !resolved) {
                 ^

TypeError: data.match is not a function
    at Socket.<anonymous> (/sites/shopify/js-buy-sdk/scripts/test-examples:68:18)
    at emitOne (events.js:101:20)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at Socket.Readable.push (_stream_readable.js:130:10)
    at Pipe.onread (net.js:542:20)
```

minasmart:
did node six change the default type of what `data` is?

yomexzo:
[This article](https://www.sitepoint.com/basics-node-js-streams/) and the [example here](https://nodejs.org/api/stream.html#stream_api_for_stream_consumers) reads
> By default the data you read from a stream is a Buffer object. If you are reading strings this may not be suitable for you. So, you can set encoding on the stream by calling Readable.setEncoding(), as shown below.

But i've not been able to figure out why it behaves so only on my computer.

minasmart:
Yeah, it looks like the type of data coming from `std*` on a child process or spawned process is a stream now, as documented here: `https://nodejs.org/api/child_process.html#child_process_child_stdout`

All good, just please adjust the title of the PR to reflect what this change is for 😄

mikkoh:
So I was running node `v5.7.0` and `npm run test-examples` ran just fine.

I ran with `v6.3.1` and it also ran fine. But yes `v6.2.2` did break... So basically that means and future versions this will probably be a non-issue.

yomexzo:
👍 @mikkoh 

🚢 ???

cc @minasmart

PR #184, opened 2016-08-02
Fix broken build after nested eslint dependency broke things.

minasmart:
We've been using babel-register to allow us to write es6 code in our build
tooling and have it run in node 0.12. By default, babel-register doesn't process
node modules. I had to override that behaviour because eslint was using `let` in
the version required by `broccoli-lint-eslint` at `v2.2.1`. After much
investigation, I was able to use the most recent `2.x` version of
`broccoli-lint-eslint`, because it referenced a `2.x` version of eslint. Eslint
2.x is supposed to support node 10 and up, but some intermediate versions (that
peer dependencies referenced) didn't follow through with that commitment.
**NOTE:** the 3.x series of eslint officially dropped support for node 12 via
[this commit](https://github.com/eslint/eslint/commit/58542e4) and we can not
update to eslint 3.x until we drop our support for node 12.

All this became apparent today because of a [flat-cache release](https://github.com/royriojas/flat-cache/commit/cd7aeed3304b839415c67434c1cd683dcdd29920)
that was fuzzy matched and added a dependency on [circular-json](https://github.com/WebReflection/circular-json)
which includes a reference to global `this` [here](https://github.com/WebReflection/circular-json/blob/0.2.0/build/circular-json.node.js#L184-L185)
which is totally broken under babel register (in node, `this === global` in the
global scope, but in babel-register's global scope `this === undefined`).

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -1,19 +1,6 @@
 /* global require, module */
 
 require('babel-register')({
-  ignore: function (fileName) {
-    var whitelistedPackages = [
-      'broccoli-lint-eslint'
-    ].join('|');
-
-    if (fileName.match(new RegExp(whitelistedPackages))) {
-      return false;
-    } else if (fileName.match(/node_modules/)) {
-      return true;
-    }
-
-    return false;
-  },
   presets: [
     require('babel-preset-es2015')
   ],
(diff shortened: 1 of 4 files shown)


lreeves:
Wow and I thought my rabbit holes sucked, nice find!

PR #185, opened 2016-08-04
Add send query method for GraphQL

minasmart:
This represents our http transport. We shouldn't have to remember how to properly talk with graph everywhere, so this function encapsulates all the things

### DIFF
--- a/src/send-query.js
+++ b/src/send-query.js
@@ -1,0 +1,12 @@
+import ajax from './ajax';
+
+export default function sendQuery(query, config) {
+  const url = `https://${config.domain}/api/graph`;
+  const headers = {
+    authorization: `Basic ${btoa(config.apiKey)}`,
+    'content-type': 'application/json'
+  };
+  const body = JSON.stringify({ query });
+
+  return ajax('post', url, { body, headers });
+}
--- a/tests/unit/api/send-query-test.js
+++ b/tests/unit/api/send-query-test.js
@@ -1,0 +1,77 @@
+import { module, test } from 'qunit';
+// import { step, resetStep } from 'shopify-buy/tests/helpers/assert-step';
+import Pretender from 'pretender';
+import sendQuery from 'shopify-buy/send-query';
+
+let pretender;
+
+const config = {
+  domain: 'krundle.com',
+  apiKey: 123
+};
+
+const hashedKey = btoa(config.apiKey);
+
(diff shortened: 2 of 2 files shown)


minasmart:
@tessalt @tanema wanna have a quick look-see?

src/send-query.js
```diff
@@ -0,0 +1,12 @@
+import ajax from './ajax';
+
+export default function sendQuery(query, config) {
+  const url = `https://${config.domain}/api/graph`;
+  const headers = {
+    authorization: `Basic ${btoa(config.apiKey)}`,
```
yomexzo: curious, will this be used in `node`?

tessalt:
lgtm

src/send-query.js
```diff
@@ -0,0 +1,12 @@
...
+    'content-type': 'application/json'
+  };
+  const body = JSON.stringify({ query });
```
tanema: Should there be some checking on the query or the config? I am not sure how public send-query so I am not sure how much it will matter.

tanema:
Small nit but otherwise 🎱

src/send-query.js
minasmart: Yep! Just like it is elsewhere using the isomorphic btoa module.

yomexzo: oh okay .. in that case, you might want to import/use isomorphic btoa right? afaik, this would break in node. No?

src/send-query.js
minasmart: `config` is validated by the config class. query validation would require query parsing, and building a query parser is a lot of code. I have tests around query generation, but I don't think I want to actually be validating queries client-side at any point. The server will return a meaningful message if it's broken.

src/send-query.js
minasmart: Nope. The [isomorphic-btoa](https://github.com/Shopify/js-buy-sdk/blob/master/src/isomorphic-btoa.js) module adds it globally. An example of its usage as it works today can be seen [here](https://github.com/Shopify/js-buy-sdk/blob/master/src/adapters/listings-adapter.js#L13)

yomexzo: hits my head ... it's imported in `shopify.js` ... sometimes, i just forget things.

yomexzo:
an extra LGTM doesn't hurt 😄

PR #186, opened 2016-08-05
Simplified Brocfile.js and added in a .babelrc file

mikkoh:
Hey I'm putting up this PR for discussions sake.

I've simplified the `Brocfile.js` where it doesn't register/setup Babel preset's and plugins. Doing this also allows for `babel-register` to be removed as a dev dep.

Thoughts? @minasmart @richgilbank @yomexzo

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["babel-preset-es2015"],
+  "plugins": ["transform-object-assign"]
+}
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -1,12 +1,3 @@
 /* global require, module */
 
-require('babel-register')({
-  presets: [
-    require('babel-preset-es2015')
-  ],
-  plugins: [
-    'transform-object-assign'
-  ]
-});
-
 module.exports = require('./js-buy-sdk-build');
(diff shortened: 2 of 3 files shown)


minasmart:
Removing register won't work on node 12

minasmart:
There's also separate babel rules for the build-lib, and src  directories. For build tooling, we don't really care what gets baked in or required. For src, we really do. And that config lives over [here](https://github.com/Shopify/js-buy-sdk/blob/master/build-lib/util/babel-config.js)

mikkoh:
@minasmart what do you mean Node 12? Stable is ~4 and experimental newest is ~6?

minasmart:
We have to support node `v0.12` up to current for building the library. We have to support ie9 for the actual SDK code.

richgilbank:
@minasmart Support for v0.12 wasn't a hard-and-fast rule… it was based on estimated version usage at a time (~6 months ago) when building the SDK happened `postInstall` rather than `prePublish` as well, so to be able to use the SDK at all, one would have to build it themselves. Since that's no longer the case, I think it's worth reevaluating whether or not it's a requirement, particularly if it simplifies the codebase. 
It's also worth noting here that LTS support for `v0.12` [ends in < 4 months](https://github.com/nodejs/LTS) so it's definitely worth at least considering dropping it in the near future.

minasmart:
@richgilbank great to know. I'd be more than happy to not have to run so many parallel builds in CI and update to a newer root version 😄

PR #188, opened 2016-08-10
Fix the ci test script to actually move the xunit report files in to place.

minasmart:
Previously, the test script was exiting at any failure. With this change, it captures test status, moves xunit files into place, then exits with the combined exit statuses of the test scripts to inform circle of any failures


lreeves:
👍

tanema:
👍

PR #194, opened 2016-08-16
Add helpers to discern types from fields

minasmart:
Adds to helper methods:
- `rawTypeForField(fieldName, schemaModuleName);`
  Returns the exact type for a field, as indicated by the graphql schema
- `typeForField(fieldName, schemaModuleName);`
  Returns the type for the field in a paginated-set aware way. Traversing connections
  and edges, and exposing the node type as the base type for the list.

The returned type data is the graphql type combined with an `isList` attribute which is extracted from the field

### DIFF
--- a/src/graph-helpers/raw-type-for-field.js
+++ b/src/graph-helpers/raw-type-for-field.js
@@ -5,7 +5,7 @@ function findInFields(property, type) {
 
   if (propertyDescriptor) {
     return {
-      name: 'Literal',
+      name: 'Scalar',
       isList: propertyDescriptor.isList
     };
   }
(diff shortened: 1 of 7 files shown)


minasmart:
@lreeves and @mikkoh wanna take a look?

src/graph-helpers/raw-type-for-field.js
```diff
@@ -0,0 +1,53 @@
+import graphSchema from 'graph/schema';
+
+function findInFields(property, type) {
+  const propertyDescriptor = type.fields[property];
+
+  if (propertyDescriptor) {
+    return {
+      name: 'Literal',
```
lreeves: Shouldn't this be String, or something that describes it's a Scalar of type String?

lreeves: Actually I'm not really sure what the name is supposed to represent.

minasmart: Done!

lreeves:
Yep LGTM!

PR #196, opened 2016-08-16
Fix build files minification

yomexzo:
@mikkoh noticed build files were not being minified while he was working on the build system.

Figured it is best to fix it now instead of waiting for the build system refactoring to be completed.

👀 @mikkoh @minasmart

### DIFF
--- a/build-lib/lib.js
+++ b/build-lib/lib.js
@@ -62,6 +62,11 @@ module.exports = function (pathConfig, env) {
     return mergeTrees([bareTree, polyfilledLibTree]);
   });
 
+  const nodeTree = funnel(sourceTree(pathConfig, 'commonjs'), {
+    srcDir: '.',
+    destDir: './node-lib'
+  });
+
   let mergedTree;
 
   if (env == 'production') {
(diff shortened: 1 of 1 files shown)


build-lib/lib.js
```diff
@@ -67,12 +67,14 @@ module.exports = function (pathConfig, env) {
...
   });
 
-  if (env.production) {
-    trees.push(uglifyJavaScript(funnel(trees, {
-      getDestinationPath: function (path) {
-        return path.replace(/\.js/, '.min.js');
-      }
-    })));
+  if (env == 'production') {
+    trees = [
```
minasmart: Won't this minify all trees? Previously, we were adding additional minified trees. This looks like it replaces all trees with the min versions.

Also, use a separate const over reassigning. This is a totally different tree at this point, and calling it out as `productionTrees` or `minifiedTrees` or something makes that separate identity clear.

yomexzo: `uglifyJavaScript` apparently returns all the files in the fed in tree along with the minified version of the `js` files.

@naming, i had a bit of a dilemma there .... changing now.

build-lib/lib.js
```diff
@@ -62,20 +62,24 @@ module.exports = function (pathConfig, env) {
...
-    })));
+    }));
+  } else {
+    mergedTree = mergeTrees(trees);
   }
 
+  const nodeTree = funnel(sourceTree(pathConfig, 'commonjs'), {
+    srcDir: '.',
+    destDir: './node-lib'
+  });
```
minasmart: Was there a reason to change the declaration order? Muddies up the diff.

yomexzo: i think it's better placed after the `tree` building business is done.

yomexzo:
are we good to merge this @minasmart, @mikkoh ?

We just deployed v0.2.2 but minified files were not added to the CDN. Also, latest non-minified has v0.2.2 while minified still has v0.2.1.

minasmart:
👍

PR #197, opened 2016-08-17
Feature/build system

mikkoh:
Sorry this is a biggie... The purpose of this PR is to make it easier for Contributors and new people joining Shopify. The build system is fully layed out in npm scripts which most JS devs are familiar with today.

In the process we've also updated the build system to produce cleaner/smaller bundles.

This pull requests:
1. Updates the build system to use npm scripts instead of Broccoli
2. `npm run build` now produces a UMD build in the dist folder instead of global, commonjs/iife, amd, etc
3. Rollup is now used to bundle the UMD build which seems to produce a cleaner output.
4. `lib/shopify.js` will now a have version number added to source. Not needed most likely but worth mentioning.
5. Folder structure has been made simpler so that the root only has folders for `src/`, `tests/`, `examples/` and `scripts/`

@minasmart @lreeves @yomexzo @richgilbank @tessalt


.rollup.config.umd.polfilled.js
```diff
@@ -0,0 +1,36 @@
...
+
+      // if false then skip sourceMap generation for CommonJS modules
+      sourceMap: false,  // Default: true
+
+      // explicitly specify unresolvable named exports
+      // (see below for more details)
+      namedExports: { './module.js': ['foo', 'bar' ] }  // Default: undefined 
+    })
+  ]
+};
```
minasmart: Fix EOF Character. Remove boilerplate.

scripts/addHeaderToSource.js
```diff
@@ -0,0 +1,32 @@
...
+const fs = require('fs');
+const getStdin = require('get-stdin');
+
+
+getStdin()
+.then(source => {
+  const templateVersion = require('./util/templateVersion');
+  const VERSION_TEMPLATE_STRING = require('./util/versionTemplateString');
+
+  const pathLicense = path.join(process.cwd(), 'LICENSE.txt');
```
minasmart: I think you can build the path and read the file into a buffer before this `then` and save doing IO on every rebuild (for the sake of watching).

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,39 @@
...
+
+const outSource = `
+${sourceModules[0]}
+(function (self) {
+;self.fetch = null;
+${sourceModules.slice(1).join('\n')}
+}(window));
+`;
+
+console.log(outSource);
```
minasmart: What's the console log for?

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,39 @@
...
+
+function getSourceFromModules(modules) {
+
+  return modules
+  .map((nameModule) => {
+    const pathModule = require.resolve(nameModule);
+
+    return fs.readFileSync(pathModule, 'utf8');
+  });
+}
```
minasmart: fix EOF Character.

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,39 @@
...
+;self.fetch = null;
+${sourceModules.slice(1).join('\n')}
+}(window));
+`;
+
+console.log(outSource);
+
+function getSourceFromModules(modules) {
+
+  return modules
```
minasmart: Is there a reason to not call in one line like `modules.map(...)`? It's kind of weird to me to have the line break there. I'm curious on your opinion though.

scripts/generateESLintTest.js
```diff
@@ -0,0 +1,42 @@
...
+const paths = glob.sync('src/**/*.js');
+
+const engine = new CLIEngine();
+const results = engine.executeOnFiles(paths).results;
+
+const header = `import { module, test } from 'qunit'`;
+const source = `${header}
+
+test('ESLint', function(assert) {
+${results.map(generateAssert).join('\n')}
```
minasmart: Does this jam them all into one test with multiple annotated assertions?

scripts/runBabel.js
```diff
@@ -0,0 +1,46 @@
...
+ * in and also set BABEL_ENV. If no data is piped in then it is expected that
+ * the second argument passed to this script will be the src to be run through
+ * babel.
+ */
+
+const babel = require('babel-core');
+const getStdin = require('get-stdin');
+
+// set the environment variable on which babel should run
+process.env.BABEL_ENV = process.env.BABEL_ENV || process.argv[ 2 ];
```
minasmart: `argv[2]`.

minasmart: unless this is a pattern specific for argv?

scripts/runBabel.js
```diff
@@ -0,0 +1,46 @@
...
+
+// read in data piped to this done app
+getStdin()
+.then(input => {
+  let code;
+
+  if(input !== '') {
+    // run babel on the piped in data
+    code = babel.transform(input, {
+      extends: process.cwd() + '/' + '.babelrc'
```
minasmart: why not a template string?

scripts/util/getVersion.js
```diff
@@ -0,0 +1,15 @@
...
+const gitTag = require('git-rev').tag;
+const gitCommit = require('git-rev').short;
+
+module.exports = function(callback) {
+  gitTag((tag) => {
+    gitCommit((commit) => {
+      callback(null, `${tag}-${commit}`);
+    });
+  });
+};
```
minasmart: EOF

scripts/util/templateVersion.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = (source, callback) => {
+  getVersion((err, version) => {
+    if(err) {
+      throw err;
+    }
+
+    callback(null, source.split(VERSION_TEMPLATE_STRING).join(version));
+  });
+};
```
minasmart: EOF

scripts/util/versionTemplateString.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * This is here so if this template string ever needs to change it can
+ * be easily replaced
+ */
+
+module.exports = '{{versionString}}';
```
minasmart: EOF

scripts/generateESLintTest.js
```diff
@@ -0,0 +1,42 @@
...
+});
+`;
+
+console.log(source);
```
tessalt: 🔥

src/polyfills.js
```diff
@@ -0,0 +1,18 @@
...
+
+let scope;
+
+// check if we're running in an eco-system that uses window
+// if it doesnt exist use global
+if (typeof window !== 'undefined') {
+  scope = window;
+} else {
+  scope = global;
+}
```
minasmart: Use the existing global module, or remove it if it's only going to be used in this one place now.

minasmart:
Seem to be some missing deps taht are breaking CI. `test-examples` hangs on my machine.

Overall, I really like this approach. I do have a small concern on the three second rebuild time though.... feels like sprockets :P.

One question about the enormous number of scripts: What are your thoughts about only sticking scripts that you would ever run on their own with `npm run` in the scripts area, and for the rest, jam them into files? I can discuss this IRL, because I think I can explain what I mean more easily that way.

scripts/build-testing-sdk.js
mikkoh: For flexibility of most of these scripts generated code is sent to stdout.

I could have done something like `process.stdout.write` but I think `console.log` is a bit friendlier.

For this script in particular we do:
`node scripts/build-testing-sdk.js > .dist-test/sdk-testing.js`

And since the script outputs to `stdout` it's very easy in the npm script to change the path in the future or path to another program eg.

`node scripts/build-testing-sdk.js > new-source-folder/new-file-name.js`

minasmart: Cool.

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,39 @@
...
+${sourceModules[0]}
+(function (self) {
+;self.fetch = null;
+${sourceModules.slice(1).join('\n')}
+}(window));
+`;
+
+console.log(outSource);
+
+function getSourceFromModules(modules) {
```
minasmart: Function declaration should go before first use.

scripts/build-testing-sdk.js
mikkoh: @minasmart I may have been chaining calls beyond map. Generally if I chain I've always dropped to a newline even for the first call to indicate I'll be chaining. In this case it's a mistake. I will fix... I've also noticed even when chaining people here don't drop to a newline so that'll have to be a muscle memory fix thing for me to figure out.

scripts/generateESLintTest.js
mikkoh: @minasmart yes it does.

scripts/runBabel.js
mikkoh: @minasmart nope... Muscle memory I'm trying to get out of. Will fix.

scripts/runBabel.js
mikkoh: Less chars. HA. Lazy... Will swap for template string. Funny thing though now looking at this line theres concatenation for no reason and should probably use `path` instead. Will fix.

scripts/build-testing-sdk.js
mikkoh: @minasmart is that a rule? I started doing this so that it's easier to walk through code. It allows for the main application logic to exist at the top of the page and supporting functions to be at the bottom. This means that if you have to walk through the code you see the major functionality first and then discover the supporting functions later. Should be a non-issue due to hoisting. But... If it's a rule with no discussion I will follow suite.

mikkoh:
@minasmart I'll bring in https://www.npmjs.com/package/selenium-standalone which will bring in Selenium via NPM.

scripts/build-testing-sdk.js
minasmart: I'm open to discussion, it's just weird for me to read. When I see a function call and don't see it defined or imported I immediately think "Where did this global come from?" but that's me.

scripts/addHeaderToSource.js
mikkoh: @minasmart I wish we could do that 😢 This script executes each time on rebuild since it's not `babel` watch thats building.

package.json
```diff
@@ -5,14 +5,56 @@
   "main": "lib/shopify.js",
   "repository": "git@github.com:Shopify/js-buy-sdk.git",
   "scripts": {
-    "build": "node ./scripts/build",
-    "prepublish": "node ./scripts/prepublish",
+    "start": "npm run dir-setup-dist-test && npm run dir-setup-dist && concurrently 'npm run watch-test' 'npm run build-test-sdk' 'http-server -p 4200 .dist-test/'",
```
yomexzo: since `start`, `test`, `test-examples`, and `examples` only use `.dist-test` ... may be we don't need `npm run dir-setup-dist`?

richgilbank:
image
😍  Nice. This should significantly help lower the barrier to entry on contributing to the SDK. 

cc @Shopify/javascript

yomexzo:
This definitely looks good.
I have a few suggestions on simplifying the main parts (`test`, `start`, `build` etc) of `package.json` tho. mostly based on commands occurring together. 

e.g `npm run build-test & npm run build-test-sdk` seems they'd always occur together. Grouping that under the same command would make their consumers shorter.

i'd share once i'm done putting my thoughts together.

.rollup.config.umd.polfilled.js
```diff
@@ -0,0 +1,21 @@
+import commonjs from 'rollup-plugin-commonjs';
```
bouk: typo in filename

package.json
```diff
@@ -5,14 +5,56 @@
...
+    "test": "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run build-test && npm run build-test-sdk && testem ci",
+    "test-examples": "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run build-test && npm run build-test-sdk && npm run cp-examples && npm run css-examples && npm run test-run-example",
+    "test-ci": "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run build-test && npm run build-test-sdk && testem ci -R xunit > tests.xml",
+    "examples": "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run cp-examples && npm run css-examples && http-server -p 4200 .dist-test/",
+    "build": "npm run dir-setup-dist && npm run build-all && npm run minify-all",
```
bouk: this makes me miss makefiles 🙈

scripts/addVersion.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
bouk: you should keep the shebang

bouk: also I don't think you need the `use strict`

minasmart: The shebang should only be in here if we ever plan on executing this script directly, which I don't think is the case. @mikkoh can correct me if I'm wrong.

As for the `use strict`, I think that's to satisfy some versions of node around the usage of `const` and `let`. If you know a way around that, please let me know. It's super annoying.

bouk: if it's not meant to be directly executed I don't think it should be in the `scripts/` directory, because that's what that's for

minasmart: Yeah, probably more `scripts/util` or somewhere that explicitly marks it as tooling, and not a program.
👍

.babelrc
```diff
@@ -0,0 +1,54 @@
+{ 
+  "env": {
+    "commonjs": {
+      "presets": ["es2015"]
```
lemonmade: Shared between all, probably fine to put it at the top level? Or does it have to go along with the `plugins` config?

.eslintrc
```diff
@@ -18,5 +19,8 @@
     "prefer-rest-params": 0,
     "prefer-reflect": 0,
     "valid-jsdoc": 0
+  },
+  "globals": {
+    "global": true
```
lemonmade: Should this just be using the node env?

.rollup.config.umd.polfilled.js
```diff
@@ -0,0 +1,21 @@
...
+
+  plugins: [
+    nodeResolve({
+      jsnext: true,
+      main: true
+    }),
+
+    commonjs({
+      include: 'node_modules/**',
+      exclude: [ 'node_modules/foo/**', 'node_modules/bar/**' ],
```
lemonmade: Supposed to be here?

package.json
```diff
@@ -5,14 +5,56 @@
...
-    "test": "node ./scripts/test",
-    "test-examples": "node scripts/test-examples",
-    "test-ci": "./scripts/ci"
+    "clean": "rimraf dist/; rimraf .dist-test/; rimraf lib/",
+    "------ Scripts used to build JS for distribution": null,
+    "build-all": "concurrently 'npm run umd' 'npm run umd-polyfilled' 'npm run commonjs'",
+    "umd": "rollup -f cjs src/shopify.js | node scripts/runBabel umd | node scripts/addHeaderToSource.js > dist/shopify-buy.umd.js",
+    "umd-polyfilled": "rollup --config .rollup.config.umd.polfilled.js | node scripts/runBabel umd | node scripts/addHeaderToSource.js > dist/shopify-buy.umd.polyfilled.js",
+    "commonjs": "cross-env BABEL_ENV=commonjs babel -d lib/ src/ && node scripts/addVersion lib/version.js",
+    "------ Scripts to compile JS for testing": null,
```
lemonmade: lol, good 'ol JSON, who needs comments

scripts/addHeaderToSource.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const license = fs.readFileSync(pathLicense, 'utf8');
+  const header = `/*\n${license}\n*/\n\n/* version: ${VERSION_TEMPLATE_STRING} */`;
+
+  // add in the header to the source
+  source = `${header}\n\n${source}`;
+
+  // now drop in the current version
+  templateVersion(source, (err, sourceVersioned) => {
```
lemonmade: This seems like a really roundabout way of getting the version into the header, if that header is the only place the version is used. Is there a reason we can't just fetch the version and embed it in the header comment?

lemonmade: Ah, I guess it's so that we can reuse this in `addVersion`. Still seems rather roundabout, I feel like there is a way we can replace the one use of `addVersion` to get around this.

scripts/addVersion.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * The purpose of this build script is to take a file on disk and template in the version
+ * number of the current build. The version of the current build is the latest git tag and commit
+ */
+
+const path = require('path');
+const fs = require('fs');
+const templateVersion = require('./util/templateVersion');
+const VERSION_TEMPLATE_STRING = require('./util/versionTemplateString');
```
lemonmade: Unused

scripts/generateESLintTest.js
```diff
@@ -0,0 +1,44 @@
+/**
```
lemonmade: Why do this instead of just running ESLint?

scripts/util/getVersion.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * This script will return a new version string which can be used in the
+ * header of the distributed js files and in source.
+ */
+
+const gitTag = require('git-rev').tag;
```
lemonmade: Is `use strict` not needed here?

.babelrc
lemonmade: Also, don't need to use the specific Rollup ES2015 config?

scripts/addHeaderToSource.js
minasmart: The version also needs to be added to the code. We send the SDK version as a header to the server so we can see how many people are on what version. We do this by putting a template string somewhere in the library like `{{version}}` and the processor drops it in.

tests/index.html
```diff
@@ -10,11 +10,21 @@
...
 
+  <script src="/sdk-testing.js"></script>
+  <!-- this needs thought but it brings in polyfills for the amd build :( -->
+  <script src="shopify-buy.umd.polyfilled.js"></script>
+  <script src="/qunit-shim.js"></script>
+  <script src="/shopify-buy.amd.js"></script>
+  <script src="/eslint-tests.js"></script>
+  <script src="/tests.js"></script>
+
+  <!--
```
lemonmade: :fire:

scripts/generateESLintTest.js
minasmart: This way, as your developing on the library, you have one place to look for output. I've found its overly optimistic to assume people will run eslint locally, and not just let CI do it. This way, you can run your tests in a browser, and see if everything is functionally good, and the code is clean. You end up with fewer people waiting on CI for linter output this way.

tests/index.html
```diff
@@ -10,11 +10,21 @@
   <div id="qunit"></div>
   <div id="qunit-fixture"></div>
 
+  <script src="/sdk-testing.js"></script>
+  <!-- this needs thought but it brings in polyfills for the amd build :( -->
+  <script src="shopify-buy.umd.polyfilled.js"></script>
+  <script src="/qunit-shim.js"></script>
+  <script src="/shopify-buy.amd.js"></script>
+  <script src="/eslint-tests.js"></script>
+  <script src="/tests.js"></script>
```
lemonmade: Always hate test runners where you need to load up specific script tags in a test HTML page, feels so weird in the era of JS modules.

scripts/generateESLintTest.js
lemonmade: I still find this really weird, and I've rarely seen stuff like this in the wild. If you are concerned with people not being able to see all the output in one place, why not just add `eslint .` as part of your `npm test`? I don't see how people can be frustrated waiting for CI to run linters and are unwilling to run a command locally that can tell them the exact ESLint output.

minasmart: One step means less developer friction. Also, `eslint .` doesn't add the output to the browsers test output. EmberJS has this as a built-in feature, and it was one of the most convenient things working with their testing infrastructure. The main thing for me is why add an additional step, and an additional place to check output, when we can do it all in one place. For me, it's a nice to have, and makes it easier to work on this project. It reduces developer friction overall, weird or not.

lemonmade: Mmk, can't say I'm the biggest fan, but if it makes it easier for your team then go for it 👍

minasmart: Having worked this way a bunch, I'd say it's pretty nice 😄

tessalt: we actually did this in brochure-land too, iirc, with jshint+mocha.

tests/index.html
tessalt: also curious about this, I couldn't find a clean way to have testem run separate top-level "suites" without using html files

package.json
perrupa: `test`, `test-examples` and `test-ci` all seem to start the same way, would it make sense to pull that out?
eg:
```json
{
  "...": null,
  "test-setup":    "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run build-test && npm run build-test-sdk",
  "test":          "npm run test-setup && testem ci",
  "test-examples": "npm run test-setup && npm run cp-examples && npm run css-examples && npm run test-run-example",
  "test-ci":       "npm run test-setup && testem ci -R xunit > tests.xml",
  "examples":      "npm run dir-setup-dist-test && npm run dir-setup-dist && npm run cp-examples && npm run css-examples && http-server -p 4200 .dist-test/",
  "build":         "npm run dir-setup-dist && npm run build-all && npm run minify-all",
}
```

mikkoh:
@minasmart I've implemented the Babel plugins to:
1. add a standard header comment
2. generate tests based on eslint
3. Drop in version numbers via `process.env.VERSION` and `babel-plugin-transform-inline-environment-variables`

If you want to see source changed for each of the changes:
### Add standard header comment
Setting up standard header for CommonJS transpiles. Only added to `src/shopify.js`:
https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/.babelrc#L6

UMD build will just add it once:
https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/.babelrc#L18

AMD will add to `src/shopify.js`:
https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/.babelrc#L30-L32

### Generate Tests Based on eslint
Here's the plugin setup for generating eslint tests:
https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/.babelrc#L40-L44

### Versioning
The following sets up version in `package.json`:
https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/package.json#L22

package.json
mikkoh: @perrupa You're definitely right. Good catch.

.eslintrc
mikkoh: @lemonmade `global` is used to shim in certain functionality if needed in Node. https://github.com/Shopify/js-buy-sdk/blob/master/src/metal/global.js

lemonmade: IMO if the only place you need that global is in the one file, it's better to add it as a global there rather than in the main linting config. As it is now, you could easily use `global` anywhere without a linting warning when what you really want is to import your `global`-determining module.

mikkoh: @lemonmade good point.

mikkoh:
Hey @minasmart I've created a branch where linting + testing are not incorporated. This would be to compare performance of watching with and without lint tests. I've implemented it so that there's a process doing eslint watch so you'll still get output in terminal for linting as files are being changed.

https://github.com/Shopify/js-buy-sdk/tree/feature/build-system%2B%2Bno-lint-tests

minasmart:
Nice! I'll take a look. Thanks @mikkoh

.gitignore
```diff
@@ -27,3 +27,7 @@ docs/src
 _site
 .sass-cache
 .jekyll-metadata
+
+# Junk
+.DS_Store
+
```
minasmart: I would just add this to your global gitignore. It doesn't need to live here, because it's a platform level concern.

.rollup.config.umd.polyfilled.js
```diff
@@ -0,0 +1,20 @@
...
+      jsnext: true,
+      main: true
+    }),
+
+    commonjs({
+      include: 'node_modules/**',
+      sourceMap: false
+    })
+  ]
+};
```
minasmart: New line at EOF

.rollup.config.umd.polyfilled.js
```diff
@@ -0,0 +1,20 @@
+import commonjs from 'rollup-plugin-commonjs';
+import nodeResolve from 'rollup-plugin-node-resolve';
+
+export default {
+  entry: 'src/shopify-polyfilled.js',
+
+  format: 'cjs',
```
minasmart: Should this not be umd?

package.json
```diff
@@ -23,49 +66,50 @@
...
-    "broccoli-merge-trees": "1.1.1",
-    "broccoli-plugin": "1.2.1",
-    "broccoli-sane-watcher": "1.1.4",
-    "broccoli-sass-source-maps": "1.7.0",
-    "broccoli-static-compiler": "0.2.2",
-    "broccoli-uglify-js": "0.1.3",
-    "broccoli-uglify-sourcemap": "1.1.1",
-    "escope": "3.3.0",
+    "concurrently": "2.2.0",
+    "core-js": "~2.4.1",
```
minasmart: Please pin this version. I'd rather not have drifting sets of functionality outside of our control.

circle.yml
```diff
@@ -4,8 +4,7 @@ machine:
   node:
     version: v5.5.0
   environment:
-    NODE_012_VERSION: v0.12.0
-    NODE_4_VERSION: v4.2.2
+    NODE_4_VERSION: v4.5.0
```
mikkoh: @minasmart @richgilbank how do you feel about this? I know in pervious discussions there was no issue moving away from v0.12.0. Doing this basically makes it so that we can use some es2015 features in Node for the build tool. If 0.12.0 is mandatory maybe we can transpile the `scripts/` dir and then run those but seems a bit yucky to do that.

minasmart: I would be totally happy moving to the current node LTS, and removing all parallelization from the build. Or at most, do two: LTS and current.

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,40 @@
+/**
```
minasmart: Any scripts that are directly executed, I would remove the `.js` and add a shebang.

scripts/build-testing-sdk.js
```diff
@@ -0,0 +1,40 @@
...
+;self.fetch = null;
+${sourceModules.slice(1).join('\n')}
+}(window));
+`;
+
+// output to stdout so that we can
+// easily modify the output/output location via cli
+console.log(outSource);
+
+function getSourceFromModules(modules) {
```
minasmart: I find it really hard to read when you're using a function before its declaration. It makes a lot more sense to me to read a procedural module as a procedure, where the actual program executes sequentially, and the end of the script is at the end.

scripts/runBabel.js
```diff
@@ -0,0 +1,48 @@
...
+if(process.env.BABEL_ENV === undefined) {
+  throw new Error('BABEL_ENV is not set. This can be set either by passing in a value to runBabel.js or set via command line');
+}
+
+// read in data piped to this done app
+getStdin()
+.then(input => {
+  let code;
+
+  if(input !== '') {
```
minasmart: For the sake of consistency with the rest of the project, (and this is totally nitpicky), but spaces after ifs. Since if isn't a function.

scripts/runBabel.js
```diff
@@ -0,0 +1,48 @@
...
+    // run babel on the piped in data
+    code = babel.transform(input, {
+      extends: path.join(process.cwd(), '.babelrc')
+    }).code;
+  } else {
+    const source = process.argv[3];
+    
+    if(source === undefined) {
```
minasmart: I wouldn't worry about type-safe falseyness. Would it be OK to just do `if (!source)`?

src/version.js
```diff
@@ -1,4 +1,4 @@
-const version = '{{versionString}}';
+const version = process.env.VERSION; // eslint-disable-line
```
minasmart: How does this work when being run on the web. Is this one of the modules you built, or a new one that's being brought in?

tests/unit/lint/lint-test.js
```diff
@@ -0,0 +1,5 @@
+import { module, test } from 'qunit'; // eslint-disable-line
+
+/**
+ * Linting CI tests will be injected here
+ */
```
minasmart: Is this working, or is it a TODO still?

.gitignore
mikkoh: Someone is bound to come on the project and commit some .DS_Store files. I think it would just save headaches to add it here instead of having to discuss get someone to fix it later.

.rollup.config.umd.polyfilled.js
mikkoh: No initially we bundle to commonjs then do the umd build from cjs. I'd like to try and see what would happen if we switch cjs to es6.

circle.yml
mikkoh: Yeah... I mean parallelization might be handy if we want to test node distributions. But now... It might not be handy other than to ensure all contributors would be able to work with it.

minasmart: Yeah, that's how it's currently used. I think I have it configured to spin up four VMs to test in different versions. Narrowing to two won't be a huge change.

.gitignore
minasmart: I think given that it's an OS specific thing, and we don't also ignore things like `thumbs.db`, I would much rather let developers manage their own environment config, and only worry about files that the project generates here.

scripts/build-testing-sdk.js
mikkoh: Hey I'd like to hear why you think this way?

I chose not to do this because editors will loose syntax highlighting in some cases it seems. For instance previously when I looked at `scripts/build` the entire script was grey.

I'm not 100% I see an advantage in doing this. Other than it would make the npm scripts shorter by removing some characters.

src/version.js
mikkoh: It's prebuilt babel plugin. [babel-plugin-transform-inline-environment-variables](https://www.npmjs.com/package/babel-plugin-transform-inline-environment-variables)

This is a pretty common thing to do in Browserify land. It's useful if you want to do something like: 
```
if (process.env.NODE_ENV === 'development') { do something just for testing }
```

Or another use case for us would be if we wanted to do something like.

```
if (process.env.ENVIRONMENT === 'node') { do something just for the node build }
```

tests/unit/lint/lint-test.js
mikkoh: This works. The code gets injected by Babel. I guess I could say:
```
Linting CI tests will be injected here by Babel
```

scripts/build-testing-sdk.js
minasmart: Oh, just because it makes it an executable script. You don't have to invoke node. It's a convenience, just to make scripts executable programs, rather than an argument to an interpreter. Removing the `.js` isn't something I'm strongly opinionated about. I have vim set up to understand shebangs and set the syntax language, but if it's a pain for other editors, leave it on 😄

src/version.js
minasmart: Oh cool! TIL 👍

tests/unit/lint/lint-test.js
minasmart: What does the plugin look for? A file with this name? Or this exact comment?

minasmart: nevermind! I found how it targets: https://github.com/Shopify/js-buy-sdk/pull/197/files/a6e1c46364b24a070ddb008e8841d31b3a71884a#diff-e56633f72ecc521128b3db6586074d2cR47

mikkoh: Yeah I tried to keep as much of the babel settings in the `.babelrc` The only thing I tried to keep in the `package.json` is input and outputs because it makes it easier to walk through what babel is building.

circle.yml
mikkoh: Yeah right now I believe it spins up 4 processes we can probably drop to 2: 4.5.0 LTS and current. It's too bad with nvm you cant do something like `nvm install latest`. I mean I guess you could do something like `nvm ls-remote | tail -1` in `setup-node`. But maybe this is out of the scope of this PR.

minasmart: I'd be all for that! But yeah, let's make that change in a smaller PR 😄

minasmart:
Just testing out, it looks like this is pretty optimistic about errors: 

image

Is that the intended behaviour? It happened when I ran without doing a fresh `npm install`

minasmart:
It also looks like selenium/wdio is still on the really verbose logging setting: 

image

At the point of those tests, the only output we care about is the passing/failing `.` characters thta pop out. Not all the webdriver actions that are being executed.

Aside from those concerns, and the other thing we talked about in person regarding moving some `npm run`able tasks into their own script files, so that the scripts entry points are really clear, and things aren't as weirdly interlinked/obfuscated, this works _really_ well and seems very performant for me, lint-test generation and all.

So just those couple things, and I think we're good to go!

mikkoh:
Ok cool @minasmart. That is very odd. The final "ok we're good to go message" will fire if -

The following files exist:
- .dist-test/sdk-testing.js
- .dist-test/tests.js

And someone is serving on tcp:4200

The only thing I can think of is that earlier on in the process we carry on even if a process fails. Hmmmm...

mikkoh:
@minasmart @yomexzo I've renamed the scripts in `package.json` to be in the form `target:action` so for instance `examples-css:build`.

https://github.com/Shopify/js-buy-sdk/blob/feature/build-system/package.json

I've also made selenium log to `selenium.log` which makes it so that selenium doesn't barf logs to the terminal. The ghost driver for some reason still logs there.

PR #201, opened 2016-08-22
Bring gh-pages into /docs folder

yomexzo:
This is a clone of `gh-pages` rebased onto master and having its contents put in the docs folder.

With this, we'd be able to serve http://shopify.github.io/js-buy-sdk/ from the docs folder

👀 @minasmart @tessalt @richgilbank @harismahmood89 

fyi, #200 (same work in another PR) was accidentally closed because i deleted the remote branch from my machine before redoing the rebase.


yomexzo:
👀 @mikkoh

yomexzo:
Just a follow up ...
I tested all our scripts and ensured they are all working and this is good to merge.

`npm run tests` - 228 tests ran & passed both on `master` and this branch.
`npm run build` - compressed the `dist` directories from both branches and confirmed both resulting zips has same file size.
`npm run test-ci` - ran as expected
`npm run examples` - i was able to use both the `cart` & `checkout` examples
`npm run test-examples` - selenium booted and 5 tests passed. Same on `master`

I have now added `npm run build-yuidoc` which handles updating of the `doc` just like we used to do on `gh-pages`.

See #202 for a list of following tasks once this is merged.

👀 @minasmart @richgilbank @mikkoh

tessalt:
github refuses to show me the entire PR but the last few commits look 👍

mikkoh:
Some notes.

1. It might be good to change the npm script "build-yuidoc" to be consistent with "doc-server". So something like: "doc-build" and place it right beside "doc-server"
2. It might be good to eventually convert scripts/yuidoc to be an npm script but it is a bit complex so keeping it as is might be the way to go.
3. `doc-server` maybe should just serve the generated `doc` folder instead of using `yuidoc` to serve. This ensures you're always reviewing the exact same content that'll be online. This might mean bringing in the Jekyll gem somehow (we can dicuss this). 
  + `jekyll serve -s docs`
4. Document the rest of `scripts/yuidoc`
5. Write documentation somewhere on how the Doc build system works

I think it'd be good to finish up #1 and then write issues for the others. Outside of that I think I'm cool with a merge. It is scarry cause it looks like a bagillion files changed but from what I can see the actual meat of the PR is:

1. .gitignore
2. .npmignore
3. package.json
4. scripts/yuidoc

Any other files we should grok @yomexzo?

minasmart:
In the readme you added, your steps are incorrect. You reference `script/yuidoc`. It's `scripts/yuidoc`. You're also missing some directory changes in the steps you mention.

yomexzo:
Thanks @mikkoh.

My intent for this PR was just to move things to the `docs` directory and ensure what was working is still working while leaving subsequent PRs for changes.

Following your feedback, i have now changed the command that generates the docs to `npm run doc-build` and the one that serves the docs to `npm run doc-serve`. `doc-serve` just runs `jekyll`. We would have to add `jekyll` as a dependency at some point like you said. However, i have updated the `readme` with more information on setting it up.

And you're right, those are pretty much the main files.

@minasmart, the `readme` is now updated and the `_config.yml` return back to the `docs` directory.

minasmart:
I think this is good? Could someone just pull it down and 🎩 it?

mikkoh:
@yomexzo I was 🎩 and I saw this

screen shot 2016-08-23 at 4 19 23 pm

mikkoh:
After doing `rm -rf node_modules && npm i` the above is fixed. All good on the 🎩

yomexzo:
Thanks @mikkoh

PR #203, opened 2016-08-25
Add class registry and graph model.

minasmart:


minasmart:
@swalkinshaw would you mind taking a look at this?

src/graph-helpers/graph-model.js
```diff
@@ -0,0 +1,15 @@
+export default class GraphModel {
+  constructor(attrs) {
+    this.attrs = attrs;
+
+    Object.keys(this.attrs).filter(key => {
+      return !(key in this);
```
swalkinshaw: I'm never sure what the best solution is here but I don't know about silently just dropping these. Could easily lead to problems debugging. What about a message through `console.warn` (think that's non-standard)?

minasmart: The whole point is that you can override a getter with your own to fill in business logic. The point is for someone to be able to `class MyThing extends GraphModel`, and it builds out the proxy getters, and allows you to write your own special ones that rely on `attrs`

swalkinshaw: Okay that makes sense 👍

swalkinshaw:
Looks good 🚀

PR #205, opened 2016-08-29
Rewrite API docs generation script into javascript from shell script

yomexzo:
This a continuation of the work at #199 but with a cleaner commit history.

It's also part of the docs road map at #202.

### What it is
This PR removes the current shell script used to generate docs and replaces it with npm script that allows you to generate docs for any of the published JS-Buy-SDK versions on your machine.

### Command
`npm run build-doc -- [--references comma-seperated-list-of-tag-or-branch-names]`

`--references` option defaults to `current` when not specified. API docs for the current version (gotten from `package.json`) is created into `/docs/api/current`.

### Example usage
`npm run build-doc -- --references master` generates API docs for the `master` branch.
`npm run build-doc -- --references v0.2.2,v0.2.1` generates API docs for the `v0.2.2` and `v0.2.1` tags.

👀 @minasmart, @mikkoh, @Shopify/javascript

Issue #202, opened 2016-08-22
Review & re-implement our docs for simplicity and encouraging contribution


yomexzo:
Considering Github Pages can now be [served from the docs directory](https://github.com/blog/2233-publish-your-project-documentation-with-github-pages). We would be moving the content of `gh-pages` into the docs directory and reformat the docs itself such that it can be browsed from within github without having to visit shopify.github.io/js-buy-sdk/.

The following will need to be done to achieve this:
- [x] Move the `gh-pages` branch content into `docs` directory within the `master` branch and ensure the docs can still be updated as it currently is on `gh-pages` branch. See #201
- [x] Implement generation of API documentation on a per-version basis such that users can still browse older docs. See #199
- [x] Implement ability to switch in between various documentation versions
- [x] Update docs content such that they are markdown files that can be easily browsed within the repo on github.
- [x] Fix `fetch`, `fetchAll`, `fetchAllPRoducts` under `ShopClient`. Currently no documentation exists for those functions but they are in the right hand pane
- [x] Write a copy and paste example which requires no prior knowledge (API keys or Product IDs). Implement into `README.md` and as first example in Docs
- [x] Update docs with information on ability to sort responses. Detailed at #175
- [x] Ensure `CONTRIBUTING.MD` reflects all changes being made to how to work with the repo

👀 @mikkoh @richgilbank @minasmart @tessalt

minasmart:
This looks great! One concern regarding:

>  Implement generation of API documentation on a per-version basis such that users can still browse older docs. See #199

Now that the docs are going to be stored in master, as human readable markdown, should we maybe just link to the tag of the old version? I'm not sure we want to keep growing the repo now that things live in master.

yomexzo:
@minasmart, I have been thinking of how best to handle that too.

We can link to the tag of the older versions just as you have said and for someone not happy with browsing the `src`, they can pull the repo, generate the docs using what's in #199 and serve the docs themselves. Of course, we'd provide instructions on how to see older docs in the docs.

This only resolves API docs tho. Not user docs that we write to compliment the API docs.

Another angle is, if we make sure to update the docs before publishing a new version, pulling down the repo, checking out a particular version and serving the docs in the version should reflect exactly the state at the time.

minasmart:
If the guides are accurate in the tagged version, that should be reflective in the source too.

If we can, at some point in the future write yuidoc to md and have jekyll bring it in, it'll be easier to read all the things, I think.

[...]

PR #206, opened 2016-08-29
Moved api docs into a subdirectory named current

yomexzo:
This PR moves all api docs into a directory named `current` within `docs/api`.

With this, our API docs will be browsed via `/api/current` on shopify.github.io/js-buy-sdk. And would allow other versions to be served. Say `/api/v0.2.0`.

However, every directory in `docs/api` except `current` is ignored so you'd have to pull the repo and manually generate the docs for other version using script added in #205.

It is part of the docs roadmap listed in #202

👀 @tessalt @mikkoh @minasmart @tanema


tessalt:
🆗

mikkoh:
LGTM. This is not a significant change but its a good one if we want to generate other versions of docs.

mikkoh:
After talking with @minasmart about documents and versioning I think we shouldn't version previous versions of docs. This would mean we don't need a current version. It's still important that at any time a developer would be able to generate and view docs from previous versions.

yomexzo:
Agreed @mikkoh. I should have closed this way back. Closing now.

PR #207, opened 2016-08-30
Only use graph types

minasmart:
I was using a mix of dasherized and canonical graph type names. The dasherized names existed more for aesthetics than anything else, but really just introduced a spot for me to make (poor) decisions. Using the canonical name everywhere removes that decision point, and makes it simpler to follow the code.

### DIFF
--- a/graph/generator.js
+++ b/graph/generator.js
@@ -57,8 +57,7 @@ function transformRelationship(field) {
     fieldName: field.name,
     isList,
     type,
-    args,
-    schemaModule: dasherize(type)
+    args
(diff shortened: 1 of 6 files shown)


minasmart:
@swalkinshaw @tessalt @lreeves can you have a look?

swalkinshaw:
Agreed with this. We decided same thing for all uses in Ruby other than dynamic method names 👍

lreeves:
👍 yup yup

tessalt:
🆗

PR #209, opened 2016-08-31
Add in a graphql deserializer

minasmart:
This adds in the deserializer chunk. It breaks fields up into scalars, relationships and lists of relationships. Instantiates a model from the class registry, and attaches relationships. Attributes that are in and of themselves models are recursively instatiated through this same function

### DIFF
--- a/src/graph-helpers/deserialize-object.js
+++ b/src/graph-helpers/deserialize-object.js
@@ -31,8 +31,6 @@ export default function deserializeObject(objectGraph, typeName, registry = new 
     return attrAcc;
   }, {});
 
-  const model = new (registry.classForType(typeName))(attrs);
-
   const relationships = relationshipDescriptors.reduce((relationshipAcc, descriptor) => {
     relationshipAcc[descriptor.fieldName] = deserializeObject(objectGraph[descriptor.fieldName], descriptor.typeName, registry);
 
@@ -52,6 +50,8 @@ export default function deserializeObject(objectGraph, typeName, registry = new 
 
     return relationshipListsAcc;
   }, {});
+
+  const model = new (registry.classForType(typeName))(attrs);
 
   Object.assign(model, relationships, relationshipLists);
 
--- a/src/graph-helpers/deserialize-object.js
+++ b/src/graph-helpers/deserialize-object.js
@@ -1,5 +1,6 @@
 import descriptorForField from './descriptor-for-field';
 import ClassRegistry from './class-registry';
+import assign from '../metal/assign';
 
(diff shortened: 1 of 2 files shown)


minasmart:
@swalkinshaw @lreeves and @tessalt would you mind taking a look?

src/graph-helpers/deserialize-object.js
```diff
@@ -0,0 +1,59 @@
...
+import ClassRegistry from './class-registry';
+
+function extractDescriptors(objectGraph, typeName) {
+  return Object.keys(objectGraph).map(fieldName => {
+    return descriptorForField(fieldName, typeName);
+  });
+}
+
+function isSingularRelationship(descriptor) {
+  return !(descriptor.typeName === 'Scalar' || descriptor.isList);
```
swalkinshaw: maybe re-use `isAttr` in these two functions?

swalkinshaw:
Code itself looks good 👍  I do wonder about some of the terminology which I think we talked about before but that's a bigger discussion probably so carry on

minasmart:
Yes. Let's set up a meeting for next week, and we can normalize/follow the spec. I'd love to get this straightened out 😄

lreeves:
Looks good. Do fixtures normally go inline or do you want to break them out? I suspect they'll get pretty gross eventualy.

swalkinshaw:
I prefer the fixtures inline if they can be kept small/self-contained like this. It's way easier to see what's going on.

minasmart:
Yeah, I always start inline. If I have a need to start sharing  them, I'll stick 'em in a module.

lreeves:
👍

minasmart:
Tests failing because phantom doesn't have object assign. Please hold.

src/graph-helpers/deserialize-object.js
minasmart: Oh! Hah! It's like a made a recyclable predicate but didn't actually do anything with it!

Will fix 😄

src/graph-helpers/deserialize-object.js
```diff
@@ -0,0 +1,60 @@
...
+function isAttr(descriptor) {
+  return descriptor.typeName === 'Scalar';
+}
+
+function isSingularRelationship(descriptor) {
+  return !(descriptor.isList || isAttr(descriptor));
+}
+
+function isRelationshipList(descriptor) {
+  return descriptor.isList && !isAttr(descriptor);
```
minasmart: Dun it!

swalkinshaw:
🚀

PR #212, opened 2016-09-03
Build System Docs

mikkoh:
@minasmart @yomexzo I wanted to keep this pull request separate from `feature/build-system` because I'm sure we can debate about wording etc for awhile. Let me know what you think.

Main changes are adding more info to `CONTRIBUTING.md` and adding `npm run info`.

I think for `CONTRIBUTING.md` the folder section is good for people just starting out but I'm not sure we need all the folders that are there now. Or if one of you would like to write docs for the important folders that would be great too.


minasmart:
This looks great! Just trying to see what's up with CI.

minasmart:
Once that passes, 🚢

PR #222, opened 2016-09-12
get image variants by size

tessalt:
I need to be able to get a specific size of image for selected variant. 

I'm really uncertain about naming here... it does the same thing as the `selectedVariantImage` getter, but needs an arg. 

@minasmart @yomexzo @mikkoh


tessalt:
nvm, gonna do this in the other lib.

PR #223, opened 2016-09-12
Initial stab at OO GraphQL query builder.

minasmart:
@swalkinshaw @lreeves would you mind having a look? I've got more to do, but this first,
then refactoring the schema parser to look more like what you're building in ruby land.


lreeves:
I think it looks pretty good! Pretty.. pretty... pretty good.

swalkinshaw:
I like it. Maps up very closely to our version right now as well 🚀

PR #225, opened 2016-09-14
use xhr in IE10

tessalt:
Since IE10 still has an `XDomainRequest` object, IE10 was making requests with XDR instead of XHR (which is supported in IE10), and they were failing for some reason that I haven't quite determined (XDR is "deprecated" in IE10 but still there). Checking for `'withCredentials' in xhr` is the recommended [solution from MS](https://blogs.msdn.microsoft.com/ie/2012/02/09/cors-for-xhr-in-ie10/) and examples page does work in IE10. 

Still works in IE9, but only if the host page is running HTTPS, which I don't think is a regression. 

@minasmart @richgilbank @mikkoh

### DIFF
--- a/examples/cart/index.js
+++ b/examples/cart/index.js
@@ -4,7 +4,7 @@ $(function() {
   ============================================================ */
   var client = ShopifyBuy.buildClient({
     apiKey: 'bf081e860bc9dc1ce0654fdfbc20892d',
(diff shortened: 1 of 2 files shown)


minasmart:
You're the 💣 @tessalt! :shipit:

PR #226, opened 2016-09-14
make console.log act like a proper function

tessalt:
I swear we fixed this before... but anyways, in IE9, `console.log` isn't a proper function and doesn't support `apply`, which is what `...arguments` desugars to. 

@minasmart @richgilbank @mikkoh


minasmart:
This fails because `bind` isn't available in phantom. It is IE9 and up though. This might help us? https://github.com/tom-james-watson/phantomjs-polyfill

minasmart:
👍

PR #228, opened 2016-09-16
Match the graph spec for schema parsing

minasmart:
We know have three main types of fields:
- scalars
- objects
- connections

This better reflects what graph actually calls things.

Also, update helper functions to support.

Also, here's an example of a destructured shop schema:
```json
{
  "name": "Shop",
  "kind": "OBJECT",
  "scalars": {
    "city": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "countryCode": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "countryName": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "currency": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "description": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "domain": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "moneyFormat": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "name": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    },
    "province": {
      "type": "String",
      "kind": "SCALAR",
      "isList": false,
      "args": []
    }
  },
  "objects": {},
  "connections": {
    "collections": {
      "type": "CollectionConnection",
      "kind": "OBJECT",
      "isList": false,
      "args": [
        "first",
        "after",
        "sortKey",
        "reverse",
        "query"
      ]
    },
    "products": {
      "type": "ProductConnection",
      "kind": "OBJECT",
      "isList": false,
      "args": [
        "first",
        "after",
        "sortKey",
        "reverse",
        "query"
      ]
    }
  },
  "fieldOf": [
    {
      "type": "QueryRoot"
    }
  ]
}

```


minasmart:
@lreeves @swalkinshaw @mikkoh wanna have a look?

graph/generator.js
```diff
@@ -14,63 +16,36 @@ function transformArgument(arg) {
...
 function getParents(typeName, typeList) {
   return typeList.filter(potentialParent => {
     if (!potentialParent.fields) {
       return false;
     }
 
     return (potentialParent.fields.filter(field => {
-      return getBaseType(field.type) === typeName;
+      return getBaseType(field.type).name === typeName;
     }).length > 0);
```
swalkinshaw: What about `some` instead of `filter` + length check? (if that function is available in this build)

tests/unit/graph/graph-test.js
```diff
@@ -85,3 +85,15 @@ test('it adds connections with pagination info', function (assert) {
     }
   }`));
 });
+
+/*
```
swalkinshaw: O_O

graph/generator.js
```diff
@@ -79,55 +54,43 @@ function getParents(typeName, typeList) {
...
     return {
       name: type.name,
-      isBuiltin: isBuiltin(type.name),
-      fields: fields.map(transformField).reduce(objectifyField, {}),
-      fieldsWithArgs: fieldsWithArgs.map(transformFieldWithArgs).reduce(objectifyField, {}),
-      relationships: relationships.map(transformRelationship).reduce(objectifyField, {}),
+      kind: type.kind,
+      scalars: scalars.map(transformField).reduce(objectifyField, {}),
+      objects: objects.map(transformField).reduce(objectifyField, {}),
+      connections: connections.map(transformField).reduce(objectifyField, {}),
```
swalkinshaw: When creating these arrays of types (`objects`, etc), `getBaseType` was already run on them. Then `transformField` ends up doing the same thing again. Is possible to store the the base type on the first pass and then skip the 2nd one?

graph/generator.js
minasmart: 👍

tests/unit/graph/graph-test.js
minasmart: 🔥

graph/generator.js
```diff
@@ -14,120 +16,83 @@ function transformArgument(arg) {
...
+  }).reduce((descriptorAcc, key) => {
+    descriptorAcc[key] = field[key];
+
+    return descriptorAcc;
+  }, {});
 
   acc[field.fieldName] = descriptor;
 
   return acc;
 }
```
swalkinshaw: Since you're reducing here, what about merging the filter into this loop?

```js
function objectifyField(acc, field) {
  const descriptor = Object.keys(field).reduce((descriptorAcc, key) => {
    if (key !== 'fieldName') {
      descriptorAcc[key] = field[key];
    }
   
    return descriptorAcc;
  }, {});

  acc[field.fieldName] = descriptor;

  return acc;
}
```

Alternatively, if `fieldName` is the only property you don't want, what about something even simpler:

```js
function objectifyField(acc, field) {
  const descriptor = field;
  delete descriptor['fieldName'];

  acc[field.fieldName] = descriptor;
  return acc;
}
```

graph/generator.js
```diff
@@ -14,120 +16,83 @@ function transformArgument(arg) {
...
   return typeList.filter(potentialParent => {
     if (!potentialParent.fields) {
       return false;
     }
 
-    return (potentialParent.fields.filter(field => {
-      return getBaseType(field.type) === typeName;
-    }).length > 0);
+    return potentialParent.fields.some(field => {
+      return getBaseType(field.type).name === typeName;
```
swalkinshaw: Is `field.baseType` available here or no?

graph/generator.js
minasmart: I really hate using merging enumeration operations. Reduce _isn't_ filter, and shouldn't be filtering. Filtering isn't technically a fold, where reduce is. Reduce also shouldn't mutate the input object, which is why I didn't do the delete.

graph/generator.js
minasmart: Nope. Those are fields on other types, not the generated field descriptors.

graph/generator.js
swalkinshaw: I didn't realize that would mutate by reference. Awful :(

👍

swalkinshaw:
LGTM 🚀

PR #229, opened 2016-09-18
specify jsnext:main entry point for es6 consumers

tessalt:
https://github.com/rollup/rollup/wiki/jsnext:main

Allows consumers using rollup to require es6 modules for tree-shakability.


minasmart:
👍

PR #230, opened 2016-09-19
Updated CDN url in docs

mikkoh:
This PR just updates the CDN link in docs.

I believe this is the only place where this needs to be changed. Might be good to think about how we could template in the `v0` portion of the url as we bump versions.

@minasmart @tessalt


docs/index.md
```diff
@@ -28,7 +28,7 @@ This tool is intended for use by developers who are experienced with JavaScript.
 ## Including the Buy SDK
 
 ```html
-<script src="http://sdks.shopifycdn.com/js-buy-sdk/latest/shopify-buy.polyfilled.globals.min.js"></script>
+<script src="http://sdks.shopifycdn.com/js-buy-sdk/v0/latest/shopify-buy.umd.polyfilled.min.js"></script>
```
yomexzo: is `v0` necessary? My understanding is `latest` = `bleeding` = `latest` 😄

minasmart: It's totally necessary. A lot of people are linking to `latest`, so we thought it'd be wise to vendor it behind the major version number. That way, no breaking API changes that are going to disrupt peoples apps. Only bug fixes and additions. 😄

minasmart:
@mikkoh there's all these links here that need to be updated / removed: 
image

docs/index.md
yomexzo: I agree there's a problem (people linking to `latest`) but i seriously doubt this is a fix. People linking to `latest` do that because they ignorantly copy and paste the link we provide or they are just lazy to find out what the most recent link are. If the *most recent* version is what's presented on the doc, they'd do the same with it.

Prepending a `major` version to `latest` will also potentially cause confusion. Or is this a normal thing with open source projects?

Except `v0` is being added to explicitly support 'bleeding per major version' and not to try to fix the problem, i don't think it's necessary.

As per fixing the problem, no way comes to mind right now ... other than ensuring the proper version is presented on the documentation.

minasmart: It's a normal thing with semantic versioning. It prevents us from pushing breaking API changes out to dozens or hundreds of apps and destroying other people's products. Since we _are_ using sem-ver, it definitely fixes part of the problem. We can't actually change the functional spec of the library without bumping the major version. We can however optimize and fix bugs, and progressively enhance as we go, and people pointing to `latest` will still get the benefit of that.

yomexzo: I definitely see the value in supporting bleeding per major version but I would totally use `/0.x.x/build-file-name.js` instead of messing with `latest` via `/v0/latest/build-file-name.js`. `0.x.x` means what we're trying to provide in *semver*.

*two $*

minasmart: Not sure what you mean by your last sentence there, but I definitely think there is a use case for providing a stable API, that can incorporate progressive bug fixes. We have explicitly versioned APIs for those that want that, but this is the same argument of wether or not we pin all our versions, or fuzzy match. In the best case scenario, we support both approaches. The cool thing about this is that most people are familiar with fuzzy matching against major version, and it's far less risky (especially for those who have opted in with `latest` without understanding the risk).

mikkoh:
@minasmart @yomexzo I've dropped in code to set version numbers for download links from `package.json`. In the process I updated the downloads page links + copy.

docs/index.md
mikkoh: For "best" CDN distributions I think it's best to look at how other projects handle this.

https://code.jquery.com/

They basically do the same thing where they have a major version number they distribute.

But... Since all versions live at the root I'm not 100% sure we need `latest/` in the url.

Could be just:
```
v0/shopify-buy.umd.js
v0.3.0/shopify-buy.umd.js
v0.3.1/shopify-buy.umd.js
v1/shopify-buy.umd.js
v1.0.0/shopify-buy.umd.js
v1.1.0/shopify-buy.umd.js
```

docs/_includes/downloads.html
```diff
@@ -1,15 +1,22 @@
-{% assign sdk_version = 'v0.1.7' %}
-{% assign module_types = "globals amd common" | split: ' ' %}
+{% capture sdk_version %}v{{ site.data.package.version }}{% endcapture %}
+{% assign module_types = "umd" | split: ' ' %}
 {% assign base_url = '//sdks.shopifycdn.com/js-buy-sdk' %}
 {% assign base_filename = 'shopify-buy' %}
 
+Currently the Javascript Buy SDK is distributed as a UMD build on the CDN. This means that you should be
```
minasmart: It should be `As of v0.3.0, the Javascript Buy ...`

docs/_data/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "uglify-js": "2.7.0",
+    "wait-on": "1.5.2",
+    "wdio-dot-reporter": "0.0.5",
+    "wdio-junit-reporter": "0.1.0",
+    "wdio-mocha-framework": "0.3.1",
+    "webdriverio": "4.0.9",
+    "whatwg-fetch": "0.10.1",
+    "yuidocjs": "0.10.0"
+  }
+}
```
minasmart: OH! This file should be a symlink. git supports that. symlink please!

docs/downloads/index.md
```diff
@@ -1,6 +1,10 @@
 ---
 layout: default
 ---
+
+{% assign versionSplit = site.data.package.version | split: "." %}
+{% assign majorVersion = versionSplit[ 0 ] %}
+
 # Downloads
 
 The SDK is available as CommonJS, AMD, and global modules. CDN-hosted versions are available on the Shopify CDN, or you can download the package from NPM or Github.
```
minasmart: Isn't this text no longer accurate?

mikkoh: Yeah I guess we could just mention there's a UMD build which will work in CommonJS, AMD or global module systems.

docs/_data/package.json
mikkoh: Sorry what should be a symlink? A symlink to the parent `node_modules` folder?

mikkoh:
Locally I merged in https://github.com/Shopify/js-buy-sdk/pull/242 to test this PR and everything seemed to be good to go for me.

docs/_includes/downloads.html
mikkoh: https://github.com/Shopify/js-buy-sdk/blob/e6fabcfe31cdff1d720d94b17ef64360c935200d/docs/_includes/downloads.html#L8

docs/downloads/index.md
mikkoh: Should be good to go
https://github.com/Shopify/js-buy-sdk/blob/docs-cdn-url-update/docs/downloads/index.md#downloads

docs/_data/package.json
mikkoh: `docs/_data/package.json` is now a symlink.

mikkoh:
@minasmart @tessalt I've merged in @yomexzo's stuff to build docs. You can now 🎩 this branch if needed. I think this might be good to go.

Let me know if this good to merge.

tessalt:
👍

Issue #175, opened 2016-07-18
Product Sort Order Within a Collection?


rgdelato:
Is there currently anyway to get the custom sort order of products within a collection (and thus be able to display products in the correct order)?

homerjam:
+1

yomexzo:
that's correct @homerjam 

however, the api exposes a `sort_by` option that currently accepts the following: `updated_at`, `best-selling`, `title-ascending`, `title-descending`, `price-descending`, `price-ascending`, `created-descending`, `created-ascending`, and `collection-default`.

This can be used like this `client.fetchQueryProducts({ sort_by: 'title-descending' })`

`collection-default` is the only one that might not be clear there and it means using the sorting order you specified in the Shopify admin collection page. If you attempt to edit your collection, you would see a section that looks like the below: 

![screen shot 2016-07-22 at 11 37 52 am](https://cloud.githubusercontent.com/assets/3056958/17062398/cd2f1ec0-5000-11e6-9b97-354847d2500e.png)

homerjam:
@yomexzo Great! Thanks for letting me know

yomexzo:
@homerjam, @rgdelato .. i missed something out. It also accepts `manual`.

Manual is the custom ordering you provide in the Shopify Admin when editing/creating your collection as shown below.

![screen shot 2016-07-22 at 12 03 41 pm](https://cloud.githubusercontent.com/assets/3056958/17063288/7c2da20e-5004-11e6-9f10-01192e09cdb7.png)

You can sort by dragging each product back and forth.

PR #233, opened 2016-09-22
Docs Content Update

mikkoh:
This pull request does the following:
- Some updates to docs content
  + Made sure there were more examples in the docs
  + Examples actually have real world ids and data not just `"1234"` for ids
  + The first example encountered in docs should be copy and pasteable
- New docs created for some models
- Re-order API docs in sidebar based on "importance" rather than sorting it alphabetically
- Added a useable example to the Readme. (might not be bad to have 2 or three examples)

Would love if @minasmart @tessalt @yomexzo you could take a look. Unfortunately right now you cannot run `npm run doc:build` currently so to preview the docs in `yuidoc` just run:
```
node_modules/.bin/yuidoc --server 8080 -T default src/
```

I don't think this PR should be merged in until we can preview the actual docs but I figured I'd put this out there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,12 @@ const shopClient = ShopifyBuy.buildClient({
 
 // fetch a product using resource id
 shopClient.fetchAllProducts()
-  .then(function (products) {
-    console.log(products);
-  })
-  .catch(function () {
-    console.log('Request failed');
-  });
+.then(function (products) {
+  console.log(products);
+})
+.catch(function () {
+  console.log('Request failed');
+});
 ```
 
 ## Documentation
--- a/README.md
+++ b/README.md
@@ -8,6 +8,28 @@ checkout.
 
 If you're just getting started, please [read the docs](http://shopify.github.io/js-buy-sdk/).
 
+## Example
+```javascript
+const shopClient = ShopifyBuy.buildClient({
+  apiKey: 'bf081e860bc9dc1ce0654fdfbc20892d',
+  appId: 6,
+  domain: 'embeds.myshopify.com'
+});
+
+// fetch a product using resource id
+shopClient.fetchAllProducts()
+  .then(function (products) {
+    console.log(products);
+  })
+  .catch(function () {
+    console.log('Request failed');
+  });
+```
+
+## Documentation
+
+For full API documentation go checkout the [API docs](http://shopify.github.io/js-buy-sdk/).
+
 ## Contributing
 For help on setting up the repo locally, building, testing, and contributing
 please see [CONTRIBUTING.md](https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md).
--- a/docs/assets/scripts/docs.js
+++ b/docs/assets/scripts/docs.js
@@ -1,4 +1,37 @@
 $(function() {
+  function filterOrderSideNavItem(pageUrl, items) {
+    var ORDER = [
+      'ShopifyBuy',
+      'ShopClient',
+      'ProductModel',
+      'ProductVariantModel',
+      'CartModel',
+      'CartLineItemModel'
+    ];
+
+    var newItems = ORDER.map(function(key) {
+      return items.find(function(docClass) {
+        return docClass.name === key;
+      });
+    })
+    .filter(function(value) {
+      return value;
+    });
+
+    return newItems;
+  }
+
+  function getHTMLForSideNavItems(pageUrl, items) {
+    return items.map(function(docClass) {
+      var activeClass = "";
+      if (pageUrl.includes(docClass.url.slice(2))) {
+        activeClass = "active";
+      }
+      return "<li><a href='" + docClass.url + "' class='docs-sub-nav__link " + activeClass + "'>" +
+        docClass.name +
+        "</a></li>";
+    });
+  }
 
   var highlightSubNavItem = function () {
     var pageUrl = window.location.href;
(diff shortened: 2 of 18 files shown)


tests/unit/models/cart-model-test.js
```diff
@@ -96,6 +96,29 @@ test('it creates a line item when you add a variant', function (assert) {
...
 
+test('it creates a line item when you add a variant', function (assert) {
+  assert.expect(3);
+
+  const done = assert.async();
+
+  const quantity = 2;
+
+  const variant = singleProductFixture.product_listing.variants[1];
+
```
minasmart: This doesn't ensure an update was made. You should validate that line items is empty to start.

mikkoh: Hmmm.. Ok. I'll take a look at the code for this test. This test was actually from `master`. I just literally just copied and pasted it to ensure the old naming of the function still worked and so did the new. (running the same tests with both functions)

mikkoh: @minasmart what I did instead was to get the initial quantity if a line item exists. Then add that to the quantity.

It just didn't feel right to test if it exists first.

minasmart: Oh, I wasn't suggesting to test if it exists. I just meant that we should assert the transition, so test that our assumptions about initial state are valid, so that we can ensure that a change has occurred

tests/unit/models/cart-model-test.js
```diff
@@ -73,20 +73,57 @@ test('it proxies sub total from the underlying cart', function (assert) {
   assert.equal(model.subtotal, cartFixture.cart.subtotal_price);
 });
 
-test('it creates a line item when you add a variant', function (assert) {
+test('it deprecates addVariants to createLineItemsFromVariants', function (assert) {
```
minasmart: Honestly, instead of duplicating the test, I would stub `createLineItemsFromVariants` and test that it's called by the deprecated method. Right now, this doesn't test that. It just tests that they have similar functionality (assuming the test bodies are the same).

README.md
```diff
@@ -8,6 +8,28 @@ checkout.
...
+```javascript
+const shopClient = ShopifyBuy.buildClient({
+  apiKey: 'bf081e860bc9dc1ce0654fdfbc20892d',
+  appId: 6,
+  domain: 'embeds.myshopify.com'
+});
+
+// fetch a product using resource id
+shopClient.fetchAllProducts()
+  .then(function (products) {
```
minasmart: For the time being, I would match the style that's in the rest of the project:

```javascript
doTheThing().then(() => {
  // ...
}).then(() => {
  // ...
});
```

docs/assets/scripts/docs.js
```diff
@@ -1,4 +1,38 @@
...
+      'ProductOptionModel',
+      'CartModel',
+      'CartLineItemModel'
+    ];
+
+    var newItems = ORDER.map(function(key) {
+      return items.find(function(docClass) {
+        return docClass.name === key;
+      });
+    })
```
minasmart: Same here (match style)

tests/unit/models/cart-model-test.js
mikkoh: So is it good now? If we take the initial quantity and ensure it has increased it should be good.

tests/unit/models/cart-model-test.js
mikkoh: How would you go about testing that `addVariants` calls `createLineItemsFromVariants`?

My thought was for the end user it doesn't really matter as long as the functionality is the same.

README.md
mikkoh: Will do... I actually prefer that style vs whats in `shopify/javascript`

mikkoh:
@yomexzo any updates on getting master functional to build and serve docs?

docs/assets/scripts/docs.js
mikkoh: Hey assuming in this case it's not adding a space after `function` anything else. Linting is totally broken on this file since it's es5.

mikkoh: This should be fixed. https://github.com/Shopify/js-buy-sdk/pull/233/files#diff-e5d789301f1d0e399ded55169fc3ccf0R13

yomexzo:
@mikkoh, #205 should fix `master`. I just rebased and attended to comments on that PR.

mikkoh:
@minasmart @jamesmacaulay I'd say this PR is ready to go.

If you want to review the docs do:
```
npm run doc:build
npm run doc:serve
```

I think maybe the only thing that might be nice to add documentation to is:
`createCart` because it sais 
> attributes representing the internal state of the cart to be persisted to localStorage.

There should be an example explaining what `attrs` you'd pass to `createCart`

Or is this something we just remove from docs because it's "complex".

@minasmart could you add in the docs for this section if you feel it's important to keep?

tests/unit/models/cart-model-test.js
mikkoh: How were you thinking of testing this @minasmart?

mikkoh:
@minasmart I added a note about sorting order to `fetchQueryProducts`. Is it possible to pass in `sort_by` to `fetchQueryCollections` or other functions?

I documented `sort_by` based on this issue:
https://github.com/Shopify/js-buy-sdk/issues/175

PR #234, opened 2016-09-29
Explicitely adding btoa and atob to window scope

mikkoh:
@tessalt noticed that `btoa` and `atob` were not being polyfilled in properly.

I checked the source of the `base64` module and it was doing some fancy stuff to try to recognize whether it was being used as a module or globally.

This does not fix IE9 issues. Will need to investigate those further.

👀 @tessalt @minasmart ?

### DIFF
--- a/src/polyfills.js
+++ b/src/polyfills.js
@@ -1,9 +1,18 @@
-import promisePoly from 'core-js/fn/promise';
 import 'whatwg-fetch/fetch.js';
-import 'Base64/base64.js';
+import promisePoly from 'core-js/fn/promise';
+import base64 from 'Base64/base64.js';
 import scope from './metal/global';
 
-// write polyfills to global scope
+// drop in polyfills from base64
+if (!scope.btoa) {
+  scope.btoa = base64.btoa;
+}
+
+if (!scope.atob) {
+  scope.atob = base64.atob;
+}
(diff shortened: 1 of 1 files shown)


src/polyfills.js
```diff
@@ -1,9 +1,18 @@
...
 
-// write polyfills to global scope
+// drop in polyfills from base64
+if (!scope.btoa) {
+  scope.btoa = base64.btoa;
+}
+
+if (!scope.atob) {
+  scope.atob = base64.atob;
+}
```
minasmart: This part looks good.

src/polyfills.js
```diff
@@ -1,9 +1,18 @@
-import promisePoly from 'core-js/fn/promise';
 import 'whatwg-fetch/fetch.js';
-import 'Base64/base64.js';
+import promisePoly from 'core-js/fn/promise';
+import base64 from 'Base64/base64.js';
```
minasmart: What was the rationale behind arbitrarily reordering imports?

tessalt:
👍

mikkoh:
I'm killing this PR since I'm not 100% sure this method would work since for some reason Babel was stripping out this code anyway. This would be fixed properly in the #235 PR.

PR #235, opened 2016-10-03
Fix/ie9

mikkoh:
This PR is not ready to be merged yet. Earlier `npm start` didn't run at all in ie9.

Now it works. However there seems to be some issues with XHR. Pinging you @tessalt cause you've previously worked on `js-buy-sdk` ie9 compatibility.

Currently when running in ie9 in Browser Stack in ie9 mode (I believe attempts to run in quirks cause the doc type for tests.html is not defined) I get the following:
screen shot 2016-10-03 at 11 40 13 am

Something of note is that the following in `src/ajax.js` will never be truthy in ie9:
```
  if (!('withCredentials' in xhr)) {
    return ie9Ajax(...arguments);
  }
```


tessalt:
@mikkoh have you isolated why this regression happened in the first place? I'm not sure that tests were passing before but im certain that XHR/XDR requests were working properly before in IE9 (though only on https).

mikkoh:
@tessalt at this point the only variances I can see:

1. Polyfills were being bundled into the tested amd modules by concatenating them to the top of the source
2. Testing in Browserstack vs VM

It currently looks like only tests related to the Cart are broken so that to me speaks that possibly everything is "working" except for the cart which maybe something that wasn't tested via examples.

tessalt:
even if it wasn't tested via examples, the cart methods all worked when I was testing buybutton-js in IE9

package.json
```diff
@@ -63,6 +63,9 @@
     "babel-plugin-rename-umd-globals": "0.0.3",
     "babel-plugin-transform-es2015-modules-amd": "6.8.0",
     "babel-plugin-transform-es2015-modules-umd": "6.12.0",
+    "babel-plugin-transform-es3-member-expression-literals": "~6.8.0",
+    "babel-plugin-transform-es3-property-literals": "~6.8.0",
+    "babel-plugin-transform-es5-property-mutators": "~6.8.0",
```
minasmart: pin to exact versions.

scripts/replaceReserved.js
```diff
@@ -0,0 +1,24 @@
+var fs = require('fs');
+var path = require('path');
```
minasmart: `const`

scripts/replaceReserved.js
```diff
@@ -0,0 +1,24 @@
...
+    contents = contents.split(replaceArgs[0]).join(replaceArgs[1]);
+  });
+
+  // output contents of replaced file
+  console.log(contents);
+
+  process.stderr.write('Finished replacing reserved syntax');
+} catch(e) {
+  process.stderr.write('Fail replacing reserved syntax');
+}
```
minasmart: This file shouldn't be necessary with the babel transforms?

tests/index.html
```diff
@@ -36,11 +36,24 @@
...
+    // It would be nice to just do `var moduleName = Object.keys(require.entries)`
+    // however IE<9 does not support Object keys so we'll need to use for(var i in require.entries)
+    // instead.
+    var modules = require.entries;
+
+    for (var moduleName in modules) {
+      // this moduleName is not defined in a proto
+      if (modules.hasOwnProperty(moduleName)) {
+        // is this module a test if so we'll require it
+        if(!!moduleName.match(/-test$/)) {
```
minasmart: There's no reason to booleanize here. The only time I'd do that is if I'm in a predicate function and I don't want to return an object. And even in that case: https://github.com/Shopify/javascript/blob/2ca4167bc12857787fad1d040ba203519d48d285/packages/shopify-codemod/docs/implicit-coercion-to-explicit.md

minasmart:
@tessalt @mikkoh The main regression took place when [these plugins](https://github.com/Shopify/js-buy-sdk/pull/235/files#diff-e56633f72ecc521128b3db6586074d2cR3) were missed in the new build system. Since IE9 broke on those reserved words, that were previously being transformed.

PR #236, opened 2016-10-04
Fix Webpack + Node + add Polyfill dist on NPM

mikkoh:
This pull request implements the `"browser"` field in `package.json`. Doing this means that `node-fetch` will not be bundled in by Webpack.

This pull request also adds a `polyfilled.js` file to the root of the module during `prepublish` which means that users should be able to do 
```
var shopify = require('shopify-buy/polyfilled');
```
To have polyfilled builds in a CommonJS ecosystem.

We should probably add some sort of testing to ensure that the Node build works. This would obviously mean using something other than qunit.

👀 @minasmart @tessalt @jamesmacaulay ?


src/ajax.js
```diff
@@ -25,10 +26,14 @@ function parseResponse(response) {
...
-  const xhr = new XMLHttpRequest();
-
-  if (!('withCredentials' in xhr)) {
-    return ie9Ajax(...arguments);
+  // we need to check that we're not running in Node
+  // before we should check if this is ie9
+  if (!isNodeLikeEnvironment()) {
+    const xhr = new XMLHttpRequest();
+
+    if (!('withCredentials' in xhr)) {
```
tessalt: did you figure out why this was returning false in ie9?

minasmart:
Looks OK from a code standpoint, but I have the same curiosity as @tessalt

src/ajax.js
mikkoh: No... I'd need to look at this in the other PR. For this PR the main focus is on getting NPM up and running.

mikkoh:
@minasmart @tessalt I had to change one of the tests.

Basically it looks like Circle doesn't run tests against git repos so when it attempts to grab a version string instead of the version number being something like:
```
v0.3.0-6d0f5e2
```
It ends up being:
```
v0.3.0-no commit
```

You can see the code for this here:
https://github.com/Shopify/js-buy-sdk/pull/236/files#diff-33c1b9c71bf5942731d3e6ca987b8a1cR12

What do you think we should do about testing Node, Webpack, other build tools?

minasmart:
Does it really not test against a repo? Or does it test against a detached head? I'd say if it's detached, then change the text to something that reflects that. If it's actually outside of a git repo, then maybe just use a hyphen or underscore so our version string doesn't have whitespace?

mikkoh:
So right now the git commit has is pulled in by:
https://www.npmjs.com/package/git-rev-sync

Which is used by:
https://www.npmjs.com/get-project-version

Which speaks to me that since `git-rev-sync` can't get any hash it's not a git repo. I can ssh in to check. I think in the end the most important thing is that the version number is there which is pulled in by `git tag` and if that fails it pulls it in from `package.json`.

Might need to add the ability to define the hash value when no commit exists for `get-project-version`. I feel `no commit` is fine since you generally wont see that anyway.

mikkoh:
Hmmm... sshd into Circle. It looks like it is a git repo. Hmmm... Looking into why `git-rev-sync` is failing.

mikkoh:
@minasmart figured why it's not getting a commit on Circle. There's a bug in `git-rev-sync`

They basically get the contents of `.git/packed-refs`
Look for the latest commit based on the branch via regex.

In the following code `refToFind` is the name of this branch `"fix/webpack+node"`
```
var packfileRegex = new RegExp('(.*) ' + refToFind);
```
Since this branches name includes a `+` the above code fails to match any commit message.

Gonna do a PR see if they can fix it. Regardless not an issue with any of our modules or this module.

If you want to see the code that breaks it's below:
https://github.com/kurttheviking/git-rev-sync/blob/master/index.js#L94-L95

tessalt:
any update on this?

tessalt:
I've tried pulling this down and when I run `npm run examples:serve` I get a 404 on shopify-buy.umd.polyfilled.js . I tried running `prepublish` first but still no luck.

minasmart:
@tessalt I made a couple changes that should help example serving. Wanna give it a go again?

tessalt:
👍  exmaples serving. I'm gonna try rebasing against fix/ie9 branch and see what happens in ie9

minasmart:
Wonderful! Thank you @tessalt

tessalt:
examples work in IE9 & 10 (over https... not sure if this is a regression in IE10, it is not in IE9) but tests still do not pass

tessalt:
If tests are passing in CI and the examples _work_ then I'd like to ship this. BuyButton.js is currently broken in IE9/10 and it would be nice if it could not be.

minasmart:
Yeah, I'm with you on that @tessalt. Let's revisit getting the test runner working in other browsers at a later date if you need this in.

tessalt:
@mikkoh are you cool to merge this and the ie9 fix?

PR #237, opened 2016-10-04
Shift to just node current and LTS for testing

minasmart:
This represents the versions of node we should be supporting (LTS and current).


minasmart:
@mikkoh @tessalt thoughts?

tessalt:
seems reasonable

mikkoh:
I think this is good 👍

PR #239, opened 2016-10-05
Add files to .npmignore

mikkoh:
There were many files added to the `js-buy-sdk` when the build tool was setup. This PR adds those files to `.npmignore`.

👀 @minasmart @tessalt ?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -13,6 +13,22 @@ tmp
 # Build artifacts
 .dist-test
 tests.xml
+.babelrc
+.eslintrc.json
+.gitignore
+.npmignore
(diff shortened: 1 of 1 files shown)


tessalt:
are we still building on postinstall? removing babelrc could impact that iirc

mikkoh:
@tessalt Building during "prepublish". Which in theory gets run on "postinstall" but I don't think "prepublish" gets called if installed as a dep. At least I'd think so since dev deps arent even installed when you install a module as a dep.

minasmart:
LGTM!

tessalt:
👍

PR #241, opened 2016-10-11
add fetch for browser to deps

tessalt:
@minasmart @mikkoh

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,10 +55,11 @@
   "license": "MIT",
   "dependencies": {
     "whatwg-fetch": "0.10.1",
-    "node-fetch": "1.3.3"
+    "node-fetch": "1.3.3",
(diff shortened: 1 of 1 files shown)


minasmart:
Approved!

tessalt:
still not convinced this solves anything

tessalt:
well it solves _some_ things

mikkoh:
Totally agree with this. Actually I was thinking really all the polyfills should go into deps not dev deps.

I believe these are the only two missing from deps:
- Base64
- core-js

minasmart:
Does this fix all the things @tessalt ?

tessalt:
@minasmart remains to be seen. I think combined with the build all the things update

minasmart:
Then let's :shipit:

PR #242, opened 2016-10-11
`npm run doc:build` generates documentation src/ to docs/api by default

mikkoh:
This pull request modifies `feature/docs` slightly by default generating docs for the current src. It completely ignores `git` for doc generation and simply generates docs when you run:
```
npm run doc:build
```

@minasmart @yomexzo

### DIFF
--- a/scripts/util/doc-builder.js
+++ b/scripts/util/doc-builder.js
@@ -88,19 +88,17 @@ function checkoutSrcDirectories() {
   });
 }
 
-function copyHeadSrcDirectory() {
-  const version = 'head';
-  const srcRelativePath = path.join(options.srcDirRelPath, version);
-  const srcFullPath = path.join(process.cwd(), srcRelativePath);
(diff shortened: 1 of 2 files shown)


scripts/util/doc-builder.js
```diff
@@ -88,19 +88,17 @@ function checkoutSrcDirectories() {
   });
 }
 
-function copyHeadSrcDirectory() {
-  const version = 'head';
-  const srcRelativePath = path.join(options.srcDirRelPath, version);
-  const srcFullPath = path.join(process.cwd(), srcRelativePath);
-  const apiRelativePath = path.join(options.apiDirRelPath, version);
+function copySrcDirectory() {
```
yomexzo: otherwise, 👍

mikkoh: Yeah the only thing is you might want to generate docs before committing to head. At some point it would be good to write a watch and build docs command and this would mean you shouldn't need to commit to create docs.

PR #243, opened 2016-10-12
build should build all the things

minasmart:
This ensures that when prepublish is run, it calls build, and build actually builds all the artifacts we care about


tessalt:
👍

PR #244, opened 2016-10-12
Fix `shopify-buy/polyfilled`

mikkoh:
A few things have been done in this PR:
1. Fixed pathing for imports in `polyfilled.js` to import from `./lib` instead of `./`
2. Renamed `metal/globals.js` to be `metal/global-vars.js`
3. `global-vars.js` exposes two functions `get` and `set` which will set and get global variables. This seems to fix the issue of `globals` not being bundled in by Babel

@minasmart could you check this PR
@tessalt could you 🎩 with `buy-button-js` + ie9 using
```
import ShopifyBuy from 'shopify-buy/polyfilled';
```


package.json
```diff
@@ -95,8 +95,8 @@
     "pretender": "0.10.1",
     "qunitjs": "1.20.0",
     "rimraf": "2.5.4",
-    "rollup": "0.34.7",
-    "rollup-plugin-commonjs": "3.3.1",
+    "rollup": "~0.36.3",
+    "rollup-plugin-commonjs": "~5.0.5",
```
minasmart: pin versions

mikkoh: Someday I'll get one of these in 😏

mikkoh: Fixed here:
https://github.com/Shopify/js-buy-sdk/blob/ea0bba50ec949f6046e7454a91169f4b182721a9/package.json#L98

minasmart: use `--save-exact` in addition to `--save`

mikkoh: Yeah... Habits. Or... We could try out Yarn for awhile ;)

mikkoh:
I've run:
`npm test`
`npm run test-example`
and manually tested for IE9 using shopify/buy-button-js and Browserstack I'm confident we're good to go so I'll be merging this in.

package.json
minasmart: I'm down with trying yarn for a while. I tried it on some of the other modules I'm working on and it's so nice.

PR #247, opened 2016-10-13
Fix/docs package.json symlink

mikkoh:
This PR drops in two things:
1. Add the ability for `doc-builder` to Copy over `package.json` into `docs/_data/package.json` so that the Liquid Template can consume it. Previously it was a symlink but it seems that `Jekyll` on Github does not support symlinks.
2. The latest `package.json` is copied over to `docs/_data/package.json`

@minasmart @tessalt


yomexzo:
Hey @mikkoh, based on my understanding, If the only reason we are including `package.json` is because of `site.data.package.version`, I think we should add the `version` key to `/docs/_config.yml` and it can be accessed with `site.version`.

I say this because if `/package.json` version changes, one would have to also update `/docs/_data/package.json` (not automated). So why not define that attribute in `/docs/_config.yml` already.

minasmart:
@yomexzo https://github.com/Shopify/js-buy-sdk/pull/247/commits/6acc8b5e09a015ea0097b2039e74336661376362 adds automated copying of that file on `docs:build`

mikkoh:
@yomexzo so currently in this PR `package.json` from root is copied to `docs/_data/package.json` for verbatim. I'd rather do that then template + modify `_config.yml`. It's just less error prone and less code to just copy it over.

The reason why I implemented copying over the file in `doc-builder` is because if you wan't to generate docs based on reference we don't want to copy over the `package.json` file.

Are we good on this PR?

minasmart:
I'm good!

yomexzo:
@minasmart, i totally missed that.

@mikkoh, totally makes sense at the need to copy.

If this is the preferred method, totally good to merge.

But IMO, I don't think it's best to commit the entire file since you only need the `.version` value and since there's a cleaner alternative. In fact, everything else in that file doesn't need to be in there even if the file has to be there. Except other keys in `package.json` needs to be exposed via `sites.data.package`

yomexzo:
Yuidoc needed a similar value, it was handled here - https://github.com/Shopify/js-buy-sdk/blob/master/scripts/util/yuidoc.js#L19

I realize this is trickier tho.

minasmart:
@yomexzo the problem with jekyll is that it can't access the file system outside of `/docs`. so we need some sort of file with this value in it. since it originates in `package.json` that made sense to me.

scripts/util/doc-builder.js
```diff
@@ -89,11 +90,18 @@ function checkoutSrcDirectories() {
   });
 }
 
+function copyPackageJSON(apiDocsMeta) {
+  console.info('info:', 'Copying over package.json to docs/_data/package.json');
+  fsExtra.copySync('package.json', options.docsPackageJSONRelPath);
```
yomexzo: if you decide you only want `version`, this can turn into something like
```
fsExtra.writeJSONSync(options.docsPackageJSONRelPath, {version: latestVersion});
```

yomexzo:
I understand @minasmart. Thanks!

scripts/util/doc-builder.js
mikkoh: 👍

PR #248, opened 2016-10-14
Updated Node versions in circle.yml

mikkoh:
Not sure if we should do this each time lts + current changes but anyway.

@minasmart @tessalt


tessalt:
🆗

PR #249, opened 2016-10-14
ie9 error messaging when not runnning on https

mikkoh:
This PR just simply adds error messaging for when the `js-buy-sdk` is not run on an https server.

@minasmart @tessalt

### DIFF
--- a/src/ie9-ajax.js
+++ b/src/ie9-ajax.js
@@ -51,15 +51,25 @@ function ie9Ajax(method, url, opts) {
       }
     };
 
-    function handleError() {
-      reject(new Error('There was an error with the XDR'));
+    function handleError(error) {
+      // we should be calling reject but in browsers below ie9 it is better to do console.error
+      reject(`There was an error with the XDR\n${JSON.stringify(error, null, '  ')}`);
     }
 
     xdr.onerror = handleError;
     xdr.ontimeout = handleError;
 
-    xdr.open(method, authToUrl(url, opts));
-    xdr.send(opts.data);
+    try {
+      xdr.open(method, authToUrl(url, opts));
+      xdr.send(opts.data);
+    } catch (error) {
+      // using trim because the error message has a bunch of white space for some reason
+      if (error.message.trim() === 'Access is denied.') {
+        reject('Please ensure Shopify JavaScript Buy SDK is run over https');
+      } else {
+        reject(`There was an error when running XDR\n${error.message}`);
+      }
+    }
   });
 }
 


src/ie9-ajax.js
```diff
@@ -58,8 +58,17 @@ function ie9Ajax(method, url, opts) {
...
 
-    xdr.open(method, authToUrl(url, opts));
-    xdr.send(opts.data);
+    try {
+      xdr.open(method, authToUrl(url, opts));
+      xdr.send(opts.data);
+    } catch (error) {
+      // using trim because the error message has a bunch of white space for some reason
+      if (error.message.trim() === 'Access is denied.') {
+        reject(new Error('Please ensure Shopify JavaScript Buy SDK is run over https'));
```
minasmart: I would use the error logger to ensure this error gets properly logged. Unless that's already happening farther up the promise chain.

mikkoh: Should this be handled by the logger also? I'm assuming not because maybe there's a case where the user could recover from it.
https://github.com/Shopify/js-buy-sdk/blob/81c4086abd1da4fa707feaba9594ab26d12a2dbb/src/ie9-ajax.js#L55

mikkoh: As looking at some examples of people using xdr and error handling. Some people might do something like this:
```
    function handleError(error) {
      reject(new Error(`There was an error with the XDR\n${JSON.stringify(error)}`));
    }
```

mikkoh: With the logger implemented the user will have the following output:
`Unhandled promise rejectionError: Please ensure Shopify JavaScript Buy SDK is run over https`

Implemented here:
https://github.com/Shopify/js-buy-sdk/blob/feature/ie9-error-messaging/src/ie9-ajax.js#L69

tessalt:
👍  agree that it's better with the logger. IE can do some weird things with exceptions (like show popup windows, or just crash)

mikkoh:
@tessalt for the error handling you wrote how do you feel about doing something like this:
```
    function handleError(error) {
      reject(new Error(`There was an error with the XDR\n${JSON.stringify(error)}`));
    }
```
Do you agree that that error should not be handled by the logger?

tessalt:
@mikkoh actually I think anything that we want users to see should probably be handled by the logger, bc of weird IE behaviours

src/ie9-ajax.js
```diff
@@ -51,15 +53,24 @@ function ie9Ajax(method, url, opts) {
...
-    xdr.send(opts.data);
+    try {
+      xdr.open(method, authToUrl(url, opts));
+      xdr.send(opts.data);
+    } catch (error) {
+      // using trim because the error message has a bunch of white space for some reason
+      if (error.message.trim() === 'Access is denied.') {
+        logger.error('Please ensure Shopify JavaScript Buy SDK is run over https');
+      } else {
+        logger.error(`There was an error when running XDR\n${error.message}`);
```
minasmart: @tessalt, not positive, but I would still generally throw the error, just so that it's clear it's not handled. We're really not handling it, we're just reporting what happened, and the logger does all the fanciness of making it clear where the message is coming from. Thoughts?

mikkoh: Ideally we'd like to do something like:
```
logger.error('Please ensure Shopify JavaScript Buy SDK is ru...
reject(new Error('Please ensure Shopify JavaScript Buy SDK is ru...
```

But it's tough because `logger.error` is equivalent to `throw new Error` which means we cannot execute statements in after `logger.error`.

Advantages I see in us using `logger` is that in theory we could start logging errors people have often and be proactive in fixing those issues. The downside of the logger is that you basically get an error saying that a promise error was not handled properly with 0 line numbers :( (unless the user handles their errors properly by adding a catch statement and reviewing stack traces)

Using `reject` has similar issues where if the user isn't handling  errors and just just a dumb `catch` statement you'll have similar issues.

mikkoh: I thought that `logger.error` would cause `reject` not to be called in the above case. However it seems that `logger.error` does not work the same as `throw new Error`.

The reason why it would be nice to use logger because we could make it so that we actually log errors from all users using the `js-buy-sdk`.

mikkoh: @minasmart are you suggesting in this case we manually drop in where the error occurred?

What are your thoughts on just using `reject(new Error`?

minasmart: I think that's what @tessalt said would crash IE and not really show much that's useful?

tessalt: so IE (I think only < 10) does this thing where the first time you ever encounter a js error on any webpage it will popup a window telling you about it, then if you _uncheck_ the checkbox in that dialog it will not show you those popups again. Many people do not uncheck this checkbox. If you click the "Show Details" button, that's what usually crashes the browser in my experience. 

I think the defaults for this behaviour changed several times during the lifetime of IE9

tessalt: (the popup itself does not _usually_ give you any useful information it just tells you an error occured)

mikkoh: Interesting. Ok... Well if it's that severe lets just `console.error` and I'll make a note why we're not doing `reject`.

mikkoh:
@tessalt @minasmart ok... Errors for `xdr` are now handled via `reject` this way developers have better ability be responsive to errors. For instance they can show UI stating they cannot use the `js-buy-sdk` currently in ie9.

osamie:
@mikkoh @minasmart is this PR still applicable or can we close?

PR #251, opened 2016-10-17
inline fragments in graph queries

jamesmacaulay:
@minasmart and @mikkoh please review

This adds inline fragments to the query builder. A few things I changed along the way:

* replaced the untyped `{ name, args, node, fieldTypeCb }` field objects with `Field` instances that know how to serialize themselves with their own `toQuery` implementation.
* added an `InlineFragment` class whose instances appear together with `Field` instances in `Graph`'s `fields` array (which should soon be renamed to `selections` or similar to reflect its more varied inhabitants).
* replaced usage of `node` as a local variable name with the more specific `selectionSet`.

### DIFF
--- a/src/graph-helpers/graph.js
+++ b/src/graph-helpers/graph.js
@@ -20,6 +20,27 @@ function formatArgs(argumentHash) {
   return ` (${join(formattedArgs)})`;
 }
 
+class Field {
+  constructor(name, args, selectionSet) {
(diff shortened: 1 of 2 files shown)


mikkoh:
Should we drop `Field` and `InlineFragment` in their own files?

jamesmacaulay:
moved work to another repo

PR #253, opened 2016-10-19
fix argument index in includes polyfill

tessalt:
@minasmart 

`arguments[1]` will return the search string which could potentially be `parseInt`able.


minasmart:
LGTM and ship if urgent. Please include a follow up PR with a test.

PR #257, opened 2016-10-27
Removed comment from README.md example

mikkoh:
Noticed the comment in the `README.md` had a comment which was not correct anymore. Previously we were querying on one product.

@minasmart @tessalt


tessalt:
👍

PR #262, opened 2016-11-04
add fd param to checkout URL

tessalt:
prevents forwarding to primary domain from cart URL

@minasmart @richgilbank @tanema @michelleyschen

### DIFF
--- a/src/models/cart-model.js
+++ b/src/models/cart-model.js
@@ -90,7 +90,7 @@ const CartModel = BaseModel.extend({
       return `${item.variant_id}:${item.quantity}`;
     });
 
-    let query = `api_key=${config.apiKey}`;
+    let query = `api_key=${config.apiKey}&_fd=0`;
 
     if (typeof ga === 'function') {
       let linkerParam;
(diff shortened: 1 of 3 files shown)


minasmart:
Can you elaborate a bit on the "why" of this?

tessalt:
@richgilbank can you give a bit of history here?

minasmart:
And also: will this have any impact on google analytics?

richgilbank:
So, we're currently using cart permalinks based on a shop's `myshopify` domain. By default, if the shop has another domain set up as primary (automatically redirect to that domain), that also applies to the cart controller, so we end up redirecting to `mydomain.com/cart` and losing data somewhere along the way. [_fd=0](https://github.com/Shopify/shopify/blob/d7f3cec874f1a31a45992fb8db86d5093d621e54/app/controllers/concerns/redirection.rb#L190-L192) basically overrides that forced redirection, so going to `foo.myshopify.com` will continue through to checkout, without getting caught redirecting to the shop's primary domain. 
Ref: https://github.com/Shopify/shopify/blob/d7f3cec874f1a31a45992fb8db86d5093d621e54/app/controllers/concerns/redirection.rb#L190-L192

This is the default behaviour in v3 buy buttons too. If anything, I think it allows google analytics to work, where crossing multiple domain barriers may not.

minasmart:
Cool. Makes sense in this case, but I'm interested in more of a breakdown of what data we're losing so that we can fix the issue in core.

Issue #260, opened 2016-11-01
x-sdk-version header undefined


tessalt:
![image](https://cloud.githubusercontent.com/assets/1789029/19899320/0ecfea46-a036-11e6-9a48-5e6d79860ed5.png)

not sure when this happened. ShopifyBuy.version is also undefined. 

@minasmart @mikkoh

minasmart:
@tessalt are you building using our babelrc, or your own?

You may need to do what we've done [here](https://github.com/Shopify/js-buy-sdk/blob/master/.babelrc#L25) and [here](https://github.com/Shopify/js-buy-sdk/blob/8f3156f9506eb56aa989e202995ebfebb6f4b2ff/docs/_data/package.json#L34) using env vars if you're building yourself. If you use the version that we bundle, or compile into `lib`, version should already be injected.

minasmart:
If you're using our ES6 code, I'm not entirely sure of the best way to inject this data. @mikkoh thoughts?

mikkoh:
Sorry I had missed this conversation. There's really no nice way to inject the code if using the es6 code in this module.

I wonder if instead of injecting `process.env.VERSION` we generate es6 source of `version.js` when `start`, `test,` or `prepublish` is called.

minasmart:
Turns out the value is also missing in the UMD build. The fix we came to was to just hard code the value, and have a test that ensures `package.json` and `version.js` are in sync.

Issue #263, opened 2016-11-06
whatwg-fetch in dev and main deps


minasmart:
Looks like fe7c7286b0b53dd3efb066b0230ea7f6aaef30a3 duplicated `whatwg-fetch` between dev and main dependencies.

Tagging @yomexzo to take a look and see if it can just live in main deps.

swalkinshaw:
Closed by https://github.com/Shopify/js-buy-sdk/pull/265/commits/805f068dcbabdb3ef9ccc520bb6452f91701271c

PR #265, opened 2016-11-07
[FIX] jsnext

mikkoh:
This PR build a jsnext build to `lib-jsnext/`.

This basically means that it's the exact same source as `src/` except that the proper version number is inlined into `version.js`.

@tessalt @minasmart


package.json
```diff
@@ -31,7 +31,8 @@
     "umd-shopify-buy:minify": "uglifyjs --compress --mangle -o dist/shopify-buy.umd.min.js -- dist/shopify-buy.umd.js",
     "umd-shopify-buy-polyfilled:build": "rollup -f cjs --config .rollup.config.umd.polyfilled.js src/shopify-polyfilled.js | node scripts/runBabel umd > dist/shopify-buy.umd.polyfilled.js",
     "umd-shopify-buy-polyfilled:minify": "uglifyjs --compress --mangle -o dist/shopify-buy.umd.polyfilled.min.js -- dist/shopify-buy.umd.polyfilled.js",
-    "commonjs:build": "cross-env BABEL_ENV=commonjs VERSION=$(get-project-version --template v{{version}}-{{commit}}) babel --ignore src/polyfilled.js -d lib/ src/",
+    "commonjs:build": "rm -rf lib; cross-env BABEL_ENV=commonjs VERSION=$(get-project-version --template v{{version}}-{{commit}}) babel --ignore src/polyfilled.js -d lib/ src/",
+    "jsnext:build": "rm -rf lib-jsnext; cross-env BABEL_ENV=jsnext VERSION=$(get-project-version --template v{{version}}-{{commit}}) babel --ignore src/polyfilled.js -d lib-jsnext/ src/",
```
mikkoh: No. We should not. These are the files which are distributed to npm and you do not not want the files to be bundled into one file but they should be bundled by the end user when using the module.

mikkoh: This is true that having a single public entry point is ideal but I do worry about other tooling.

Besides a single public entry point do you see any other advantages?

minasmart: Optimized bundle size. We do tree shaking. Most current tooling stacks (browserify, web pack, gulp) don't. Removes (or at least centralizes) weird import/export behavior we've seen in other issues. Narrows support footprint because we control much more of what we expose.

mikkoh: Maybe we should disable tree shaking and see how big of an optimization tree shaking brings. Because we don't have that many external dependencies I wonder what the impact would be.

The weird export/import issue's we've had have been in our own codebase/transpilation and not in consumption of the `js-buy-sdk`. This being said people can do "bad" things currently and yes in theory it does narrow the support footprint.

Doing this will definitely be a major bump as it could break any module that may be importing something other than `shopify.js`.

I'm trying to think of other reasons to not do this but I can't find good reasons so let's go with it and revert later if need be. @tessalt do you see any issues?

tessalt: wait what would the single entry point be if we have a polyfilled and unpolyfilled version?

minasmart: Ideally we'd generate two. One not poly filled, one with. Just two separate rollup commands. Should be fairly trivial to implement.

mikkoh: The module would be consumed in the following ways:
```
import ShopifyBuy from 'shopify-buy';
import ShopifyBuyPolyFilled from 'shopify-buy/polyfilled';
```
If using `jsnext`
```
import ShopifyBuy from 'shopify-buy'; // uses es2015 export
// or the equivalent but the above resolves via `"jsnext:main": "es/shopify.js",`
import ShopifyBuy from 'shopify-buy/es/shopify-buy'; 

// then to consume the e6 version that is polyfilled
import ShopifyBuy from 'shopify-buy/es/polyfilled';
```

minasmart: Just to add to this, we should probably specify "jsnext:main" in the package.json

mikkoh: @minasmart it currently is specified but one problem is that, that build did not inject the version number. So this PR should fix that.

minasmart: Ah... cool. Yes. That makes sense to my disease addled brain now.

mikkoh: Go 😴 don't force yourself to review this stuff. `buy-button-js` is working and there are no immediate fires.

mikkoh:
@tessalt @minasmart the es5 commonjs export and jsnext exports are now being bundled into one file. Also the `jsnext` is bundled into a folder called `es`. We're now also bundling a polyfilled jsnext build.

mikkoh:
Just noticed that version was not being injected properly. Fixing.

mikkoh:
@tessalt @minasmart `version.js` should have `const version = process.env.VERSION;` replaced with a proper version number now.

Just noticed this PR would resolve:
#260 
#263

mikkoh:
@minasmart I've created an issue around documenting this in #270 I'm just going to 🎩 once with Rollup and `jsnext:main` before merging and publishing to npm.

minasmart:
Wonderful! Thank you!

PR #266, opened 2016-11-07
Moved guid to it's own file

mikkoh:
For some reason `set-guid-for.js` was not being transpiled properly for the es5 CommonJS output. 

The transpiled code looked like this:
```
exports['default'] = setGuidFor;
exports.GUID_KEY = GUID_KEY;
```

It should have looked something like this:
```
exports['default'] = setGuidFor;
exports.GUID_KEY = GUID_KEY;
module.exports = exports['default'];
module.exports.GUID_KEY = GUID_KEY;
```

Moving `GUID_KEY` to it's own file `../metal/guid-key` seems to resolve this issue.

@tessalt @minasmart

### DIFF
--- a/src/adapters/local-storage-adapter.js
+++ b/src/adapters/local-storage-adapter.js
@@ -1,7 +1,7 @@
 import CoreObject from '../metal/core-object';
 import setGuidFor from '../metal/set-guid-for';
 import Store from '../store';
-import { GUID_KEY } from '../metal/set-guid-for';
+import GUID_KEY from '../metal/guid-key';
 
 const LocalStorageAdapter = CoreObject.extend({
   constructor() {
--- a/src/metal/guid-key.js
+++ b/src/metal/guid-key.js
@@ -1,0 +1,1 @@
+export default 'shopify-buy-uuid';
--- a/src/metal/set-guid-for.js
+++ b/src/metal/set-guid-for.js
@@ -1,6 +1,5 @@
 /* eslint no-undefined: 0 complexity: 0 */
-
-const GUID_KEY = 'shopify-buy-uuid';
+import GUID_KEY from './guid-key';
 
 const GUID_PREFIX = `shopify-buy.${Date.now()}`;
 
@@ -89,4 +88,3 @@ function setGuidFor(obj) {
 }
 
 export default setGuidFor;
-export { GUID_KEY };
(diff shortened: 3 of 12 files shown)


minasmart:
Was this the only file being transformed this way?

mikkoh:
@minasmart I looked at a bunch of files and it did look like it. But I will continue investigating to see if other files are doing this.

minasmart:
Was the root cause of this ever discovered? Are there any assurances that this won't happen again on another arbitrary file?

mikkoh:
@minasmart I still need to investigate. I was in meetings all afternoon.

mikkoh:
product-model.js does the same

```
exports['default'] = ProductModel;
exports.NO_IMAGE_URI = NO_IMAGE_URI;
```

And this is what its usage looks like.
listings-serializer.js
```
var _productModel = require('../models/product-model');

var _productModel2 = _interopRequireDefault(_productModel);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }

models: {
  products: _productModel2['default']
},
```

shopify.js
```
var _productModel = require('./models/product-model');

var Shopify = {
  NO_IMAGE_URI: _productModel.NO_IMAGE_URI
};
```

All of this to me looks correct. I'm unsure why guid was failing for `buy-button-js`.

minasmart:
That's why I'm concerned. Im not yet convinced this is a solved issue, or that this fix did what we think it did. I should be in tomorrow. I'd like to revisit this and see what happens when we compile a single artifact. I feel like that may solve a lot of resolution-based weirdness.

mikkoh:
@minasmart  product-model.js does the same (only other file)

```
exports['default'] = ProductModel;
exports.NO_IMAGE_URI = NO_IMAGE_URI;
```

And this is what its usage looks like.
listings-serializer.js
```
var _productModel = require('../models/product-model');

var _productModel2 = _interopRequireDefault(_productModel);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }

models: {
  products: _productModel2['default']
},
```

shopify.js
```
var _productModel = require('./models/product-model');

var Shopify = {
  NO_IMAGE_URI: _productModel.NO_IMAGE_URI
};
```

All of this to me looks correct. @tessalt can you confirm for `buy-button-js` this was basically happening:
```
var _setGUIDFor = require('./metal/set-guid-for');
var _setGUIDFor2 = _interopRequireDefault(_setGUIDFor);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }

// the following is where it was failing where it was trying to access `GUID_KEY` on `default` 
// and not `_setGUIDFor2.GUID_KEY`
const GUID_KEY = _setGUIDFor2['default'].GUID_KEY;
```

PR #268, opened 2016-11-10
changed myShopifyDomain to domain

glumb:
Notice: myShopifyDomain is deprecated, please use domain and provide the full shop domain.


minasmart:
domain represents the full domain to the shop. So in this case `embeds.myshopify.com`

minasmart:
Thank you! 😄

PR #272, opened 2016-11-16
[FIX] Update docs to clearly differentiate example functions from js-buy-sdk functions

mikkoh:
This pull request attempts to address an issue that we received an email about.

From the email:
```
http://shopify.github.io/js-buy-sdk/examples/ - Product Add-To-Cart example - references an addVariantToCart method but no such method documented in the Reference
```

To clearly differentiate between functions in the `js-buy-sdk` and the example application I've created link outs to the example application source.

@minasmart @tessalt

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -91,7 +91,7 @@ var selectedVariantImage = product.selectedVariantImage;
 var currentOptions = product.options;
 ```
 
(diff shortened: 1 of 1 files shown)


tessalt:
cc @harisaurus since I think he made the example

minasmart:
👍

PR #280, opened 2016-12-13
Use access_token param instead of api_key for checkout

Matt888:
This changes the URL param passed to permalink checkout.  (see https://github.com/Shopify/shopify/pull/94780)

In the future with this project we should try to rename the internal use of `apiKey` and start calling it an `accessToken`.  This is a little outside the scope of my work, which was to just get all consumers of the checkout to use the right param on the URL.

I'm not sure how I can 🎩  this project, anyone want to help me out? :)

//review @yomexzo @minasmart

### DIFF
--- a/src/models/cart-model.js
+++ b/src/models/cart-model.js
@@ -96,7 +96,7 @@ const CartModel = BaseModel.extend({
       return `${item.variant_id}:${item.quantity}`;
     });
 
-    let query = `api_key=${config.apiKey}&_fd=0`;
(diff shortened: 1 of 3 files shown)


minasmart:
This is great!

Please go ahead and merge, and also cut a patch release. `npm version patch && git push --all && npm publish`, followed by publishing to the CDN on shipit.

One thing though: The old query param still has to work for existing clients.

Matt888:
Yeah I'm monitoring the usage of `api_key` in splunk and it will keep working for the foreseeable future.

PR #285, opened 2017-01-16
[GraphQL] This builds like the other graphql js projects now, and it's way simpler

minasmart:
Kickin' off the graphql build tooling.


minasmart:
@swalkinshaw @mikkoh please review.

CC: @tessalt @jzjiang, it begins!

GraphQL sources are to be based off of `refactor-to-graphql`. This branch will eventually turn into `v1.0.0-beta`.

For the time being, this moves our tests to mocha, and to shopify javascript for linting. Builds happen through rollup and babel. All linting is taken care of through a test generator that moves the linting output into a set of mocha tests (one per file).

This PR just carried over the config as a proof of concept. Work left to do will be:

- Depend on `shopify/graphql-js-client`
- Create some client initialization
- Create & register new model types for the models in our schema.

#### Nice to haves for initial release:
- Sane default field sets, but enough access to request specific fields.
- Direct GraphQL query access.
- Batching of multiple queries and resolving multiple models.

circle.yml
```diff
@@ -1,26 +1,9 @@
...
-
 machine:
   node:
-    version: v5.5.0
-  environment:
-    NODE_LTS: v4.6.0
-    NODE_CURRENT: v6.8.0
-    PHANTOMJS_VERSION: 2.0.0
-
+    version: v6.7.0
```
swalkinshaw: Can you try `v7.4.0` to see if current works?

minasmart: does it work for you?

.babelrc
```diff
@@ -1,89 +1,21 @@
-{ 
-  "plugins": [
-    "transform-es5-property-mutators",
-    "transform-es3-member-expression-literals",
-    "transform-es3-property-literals"
+{
+  "presets": [
+    "es2015-rollup"
```
mikkoh: What does `es2015-rollup` have that `shopify/web` and `shopify/node` don't? If there's something crucial it would be good to review adding to `shopify`

package.json
```diff
@@ -6,123 +6,73 @@
...
   },
   "author": "Shopify Inc.",
   "license": "MIT",
-  "dependencies": {
-    "whatwg-fetch": "0.10.1",
-    "node-fetch": "1.3.3",
-    "core-js": "2.4.1",
-    "Base64": "0.3.0"
-  },
+  "dependencies": {},
```
mikkoh: Are deps going to be re-evaluated? Or is the goal to have 0 deps?

.babelrc
minasmart: settings to skip module compilation so rollup works.

package.json
minasmart: For now, I don't need any, because rollup will bundle them in where used. As real work actually starts happening in here, it may change.

minasmart: To be clear, it's not a goal. Just incidental.

.babelrc
mikkoh: Oh I thought
```
        ["shopify/node", {
          "modules": false
        }]
```
Just did this...

minasmart: I think there might be more than that now. I'd rather let rollup handle the confit specifics of rollup though.

PR #288, opened 2017-01-17
add image model

tessalt:
so we kind of have an issue in buy button land. 
Sometimes products have images that are not attached to variants. Sometimes we like to show these images at specific sizes like "pico" or "grande". Currently, you can only access these size-specific srcs through the variant model. 

I propose an `Image` model. It's slightly gross because i want the API to be backwards compatible, but I think it works? @minasmart @mikkoh @harisaurus

### DIFF
--- a/src/models/image-model.js
+++ b/src/models/image-model.js
@@ -1,5 +1,18 @@
 import CoreObject from '../metal/core-object';
 
+const variants = [
+  { name: 'pico', dimension: '16x16' },
+  { name: 'icon', dimension: '32x32' },
+  { name: 'thumb', dimension: '50x50' },
+  { name: 'small', dimension: '100x100' },
+  { name: 'compact', dimension: '160x160' },
+  { name: 'medium', dimension: '240x240' },
+  { name: 'large', dimension: '480x480' },
+  { name: 'grande', dimension: '600x600' },
+  { name: '1024x1024', dimension: '1024x1024' },
+  { name: '2048x2048', dimension: '2048x2048' }
+];
+
 const ImageModel = CoreObject.extend({
   constructor(attrs) {
     Object.keys(attrs).forEach(key => {
@@ -8,7 +21,7 @@ const ImageModel = CoreObject.extend({
   },
 
   /**
-    * Image variants available for a variant. An example value of `imageVariant`:
+    * Image variants available for an image. An example value of `imageVariant`:
     * ```
     * [
     *   {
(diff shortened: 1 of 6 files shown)


minasmart:
Is this 100% backward compatible?

tessalt:
I'm _pretty sure_. I'm gonna have to write tests lol. And will point buy-button-js at it to see if anything breaks. But anything that previously pointed to selectedvariant.image or variant.imageVariants[n] will now hit an instance of ImageModel which should have instance variables matching all the attributes (which is why i copied keys rather than making getters).

src/models/product-variant-model.js
```diff
@@ -150,7 +150,13 @@ const ProductVariantModel = BaseModel.extend({
       return image.variant_ids.indexOf(id) !== -1;
     })[0];
 
-    return (variantImage || primaryImage);
+    const productImage = variantImage || primaryImage;
```
minasmart: I would just call this `image`. `productImage` in this context is sorta misleading, and `image` is what we're returning.

tests/unit/models/product-model-test.js
```diff
@@ -68,7 +68,7 @@ test('it proxies attrs for most commonly used props', function (assert) {
 
   assert.equal(model.id, singleProductFixture.product_listing.product_id);
   assert.equal(model.title, singleProductFixture.product_listing.title);
-  assert.deepEqual(model.images, singleProductFixture.product_listing.images);
+  assert.deepEqual(model.images[0].id, singleProductFixture.product_listing.images[0].id);
```
minasmart: probably don't need `deepEqual`

tests/unit/models/product-variant-test.js
```diff
@@ -144,11 +144,11 @@ test('it returns the image variants for the variant', function (assert) {
 test('it returns the image for the variant', function (assert) {
   assert.expect(2);
 
-  assert.deepEqual(model.image, baseAttrs.product.images[1]);
+  assert.deepEqual(model.image.id, baseAttrs.product.images[1].id);
```
minasmart: `assert.equal`

tests/unit/models/product-variant-test.js
```diff
@@ -144,11 +144,11 @@ test('it returns the image variants for the variant', function (assert) {
...
 
   model.attrs.variant.id = 'abc123';
 
-  assert.deepEqual(model.image, baseAttrs.product.images[0], 'the first image is default when no id matches');
+  assert.deepEqual(model.image.id, baseAttrs.product.images[0].id, 'the first image is default when no id matches');
```
minasmart: also `assert.equal`

src/models/image-model.js
```diff
@@ -0,0 +1,57 @@
...
+    *   },
+    *   {
+    *     "name": "compact",
+    *     "dimensions": "160x160",
+    *     "src": "https://cdn.shopify.com/s/files/1/1019/0495/products/alien_146ef7c1-26e9-4e96-96e6-9d37128d0005_compact.jpg?v=1469046423"
+    *   }
+    * ]
+    * ```
+    *
+    * @property imageVariant
```
minasmart: shouldn't this be `@attribute variants`?

tessalt: :p yes

PR #289, opened 2017-01-18
[FEATURE] Initial version of Client

jzjiang:
This is an initial version of the client using graphql-js-client.

### DIFF
--- a/src-graphql/base64.js
+++ b/src-graphql/base64.js
@@ -1,7 +1,9 @@
 export default function base64(str) {
   if (typeof btoa === 'function') {
     return btoa(str);
-  } else {
+  } else if (typeof Buffer !== 'undefined') { // eslint-disable-line no-negated-condition
     return Buffer.from(str).toString('base64');
+  } else {
+    throw new Error('No native way to base64 encode');
   }
 }
--- a/src-graphql/base64.js
+++ b/src-graphql/base64.js
@@ -1,9 +1,0 @@
-export default function base64(str) {
-  if (typeof btoa === 'function') {
-    return btoa(str);
-  } else if (typeof Buffer !== 'undefined') { // eslint-disable-line no-negated-condition
-    return Buffer.from(str).toString('base64');
-  } else {
-    throw new Error('No native way to base64 encode');
-  }
-}
--- a/src-graphql/base64.js
+++ b/src-graphql/base64.js
@@ -1,0 +1,7 @@
+export default function base64(str) {
+  if (typeof btoa === 'function') {
+    return btoa(str);
+  } else {
+    return Buffer.from(str).toString('base64');
+  }
+}
--- a/src-graphql/client.js
+++ b/src-graphql/client.js
@@ -8,6 +8,5 @@ export default class Client {
     const authHeader = `Authorization: Basic ${base64(config.storefrontAccessToken)}`;
 
     this.graphQLClient = new GraphQLClientClass(types, apiUrl, {headers: authHeader});
-
   }
 }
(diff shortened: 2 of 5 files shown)


src-graphql/client.js
```diff
@@ -0,0 +1,13 @@
+import GraphQLJSClient from '@shopify/graphql-js-client';
```
swalkinshaw: I think we should import this as `GraphQLClient`. @minasmart thoughts?

src-graphql/client.js
```diff
@@ -0,0 +1,13 @@
...
+import types from '../types';
+import base64 from './base64';
+
+export default class Client {
+  constructor(config, GraphQLClientClass = GraphQLJSClient) {
+    const apiUrl = `https://${config.domain}/api/graphql`;
+    const authHeader = `Authorization: Basic ${base64(config.storefrontAccessToken)}`;
+
+    this.graphQLClient = new GraphQLClientClass(types, apiUrl, {headers: authHeader});
+
```
swalkinshaw: 🔥  extra line here

src-graphql/base64.js
```diff
@@ -0,0 +1,7 @@
+export default function base64(str) {
+  if (typeof btoa === 'function') {
+    return btoa(str);
+  } else {
+    return Buffer.from(str).toString('base64');
```
mikkoh: I believe since we shim in `btoa` you can just safely use `btoa` in Node land
https://github.com/Shopify/js-buy-sdk/blob/f958d6b79873b5770413ceceef0fe84acd525794/src/polyfills.js

src-graphql/client.js
```diff
@@ -0,0 +1,13 @@
+import GraphQLJSClient from '@shopify/graphql-js-client';
+import types from '../types';
+import base64 from './base64';
+
+export default class Client {
+  constructor(config, GraphQLClientClass = GraphQLJSClient) {
+    const apiUrl = `https://${config.domain}/api/graphql`;
+    const authHeader = `Authorization: Basic ${base64(config.storefrontAccessToken)}`;
```
mikkoh: Should the token already be `base64` encoded at this point?

src-graphql/base64.js
mikkoh: I believe since we shim in `btoa` you can just safely use `btoa` in Node land
https://github.com/Shopify/js-buy-sdk/blob/f958d6b79873b5770413ceceef0fe84acd525794/src/polyfills.js

src-graphql/base64.js
```diff
@@ -0,0 +1,7 @@
+export default function base64(str) {
+  if (typeof btoa === 'function') {
+    return btoa(str);
+  } else {
```
mikkoh: We should check whether `Buffer` exists also. For instance I maybe using an un-shimmed version of the `js-buy-sk` in the Browser on an older browser. In this case we'd receive an error stating `Buffer` is not defined.

The else statement should guide the developer to use the shimmed version of the `js-buy-sdk`

src-graphql/base64.js
```diff
@@ -1,7 +1,9 @@
 export default function base64(str) {
   if (typeof btoa === 'function') {
     return btoa(str);
-  } else {
+  } else if (typeof Buffer !== 'undefined') { // eslint-disable-line no-negated-condition
```
minasmart: You can test that it's a constructor instead just not being undefined. `typeof Buffer === 'function=`, then you don't need to turn off the lint rule.

src-graphql/client.js
swalkinshaw: Seems like the "js" is an artifact of the repo name since we have multiple clients. Once we're in the context of JS I don't think it's needed for an import.

minasmart: How do you feel if, for the time being, we keep it just for the sake of being consistent, then figure out what to actually name this library?

swalkinshaw: Completely fine 👍  I assume a lot is going to change anyway

PR #291, opened 2017-01-19
[FEATURE] Fetch resources using graphql-js-client

jzjiang:
This PR adds the ability for the Client to fetch resources (products and collections) using the graphql-js-client. Supports dynamic field selection through #297.

### Examples
**Fetching a product by id**
```javascript
client.fetchProduct('7857989384', productQuery(['id', 'handle', 'title', 'updatedAt', ['images', imageConnectionQuery(['id', 'src'])], ['options', optionQuery(['name'])], ['variants', variantConnectionQuery(['price', 'weight'])]]));
```
**Fetching a collection by id** 
```javascript
client.fetchCollection('369312584', collectionQuery(['title', 'updatedAt', ['image', imageQuery(['src'])]]));
```

Pagination will be added in #300.


package.json
```diff
@@ -48,16 +48,18 @@
     "eslint-test-generator": "1.0.5",
     "expect.js": "0.3.1",
     "fb-watchman": "1.9.0",
+    "fetch-mock": "^5.8.1",
```
swalkinshaw: All other packages have a specific version number. I'm guessing this was done via `npm install --save-dev <package>` which uses `^`.

@minasmart should we stick to strict versions?

src-graphql/client.js
```diff
@@ -0,0 +1,78 @@
...
+        Accept: 'application/json',
+        Authorization: authHeader
+      }
+    });
+  }
+
+  fetchAllProducts() {
+    const query = this.graphQLClient.query((root) => {
+      root.add('shop', (shop) => {
+        shop.addConnection('products', {args: {first: 10}}, (products) => {
```
swalkinshaw: Are we going to make `first` configurable/overridable? I'm fine if it's in a separate PR though.

src-graphql/client.js
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    return this.graphQLClient.send(query).then((response) => {
+      return response.model.shop.products;
+    });
+  }
+
+  fetchProduct(id) {
+    const query = this.graphQLClient.query((root) => {
+      root.add('product', {args: {id: `gid://shopify/Product/${id}`}}, (product) => {
```
swalkinshaw: We should probably add a utility function to generate GIDs at some point.

test-graphql/client-test.js
```diff
@@ -0,0 +1,217 @@
...
+import types from '../types';
+import base64 from '../src-graphql/base64';
+import {singleProductFixture, multipleProductsFixture} from '../fixtures/product-fixture';
+import {fetchMock} from './fetch-mock-node';
+
+suite('client-test', () => {
+  const querySplitter = /[\s,]+/;
+
+  function tokens(queryString) {
+    return queryString.split(querySplitter).filter((token) => Boolean(token));
```
swalkinshaw: @minasmart maybe we can expose this from `graphql-js-client` if we're going to be using it outside of it as well?

test-graphql/client-test.js
```diff
@@ -0,0 +1,217 @@
...
+
+    const client = new Client(config);
+
+    fetchMock.post('https://multiple-products.myshopify.com/api/graphql', multipleProductsFixture);
+
+    return client.fetchAllProducts().then((products) => {
+      assert.ok(Array.isArray(products), 'products is an array');
+      assert.equal(products.length, 2, 'there are two products');
+
+      const passedArgs = fetchMock.lastCall('https://multiple-products.myshopify.com/api/graphql');
```
swalkinshaw: Could destructure this if all you're using is the body:

```js
const [_url, {body}] = fetchMock.lastCall('https://multiple-products.myshopify.com/api/graphql');
```

Just a stylistic thing though (also eslint will complain without a config change. See https://github.com/Shopify/graphql-js-client/pull/49/files#diff-df39304d828831c44a2b9f38cd45289cR7

package.json
minasmart: Yeah, always `--save-exact`. Pin versions please 😄

src-graphql/client.js
minasmart: I'd say separate PR. We also need to figure out how to make the list of attrs a field that can be passed in. That all  comes later.

src-graphql/client.js
minasmart: 👍 

Something like `gidFor(model)`, which can grab it's type and id in pretty deterministic ways.

test-graphql/client-test.js
minasmart: I'd be down for that. James looked into using `graphql-js` for query parsing and comparison. I still think there's some hope there, but it's gonna take some work.

test-graphql/client-test.js
minasmart: The unused vars rule should never be globally enabled though, because we generally don't want to adopt that pattern. It's just a side effect of writing tests that consume data in a way a real user of the SDK never would.

test-graphql/client-test.js
swalkinshaw: Yeah even aside from the "query splitting", I do think it's useful to ensure we're generating valid queries but that should be done in `graphql-js-client` and not here I think. Any consumer of the client should be able to trust its queries.

test-graphql/client-test.js
swalkinshaw: It's not a side effect in this context since this isn't SDK usage, it's `fetchMock` usage. In this case it returns an array and we may not care about the first item (the URL). Destructuring makes the code simpler/more terse (to me at least) and that rule enables it.

Agreed no-unused-vars is a good thing for app code, but it's super annoying to selectively disable throughout tests. Any war to have different rules for app vs test code (which should generally be avoided but there are a few things like this).

minasmart: maybe just add a `.eslintrc.json` in the test root with this rule? eslint configs are composable like that 😄

swalkinshaw: Oh nice I'll do that for my other PR which had this changed globally.

fixtures/product-fixture.js
```diff
@@ -0,0 +1,356 @@
...
+              "price": "0.00",
+              "weight": 0
+            }
+          }
+        ]
+      }
+    }
+  }
+};
+export const multipleProductsFixture = {
```
minasmart: This is probably more appropriately `shop-with-products-fixture.js`

package.json
```diff
@@ -48,16 +48,18 @@
...
     "fs-extra": "1.0.0",
     "http-server": "0.9.0",
     "jquery": "3.1.1",
     "minimist": "1.2.0",
     "mocha": "3.1.2",
-    "node-fetch": "1.3.3",
+    "node-fetch": "^1.6.3",
```
minasmart: exact

package.json
```diff
@@ -48,16 +48,18 @@
...
     "node-jq": "0.3.13",
     "rollup": "0.37.0",
     "rollup-plugin-babel": "2.7.1",
     "rollup-plugin-commonjs": "5.0.5",
+    "rollup-plugin-json": "^2.1.0",
```
minasmart: exact

scripts-graphql/rollup-tests.js
```diff
@@ -43,6 +45,10 @@ function envRollupInfo({browser, withDependencyTracking}) {
 
   if (browser) {
     plugins.unshift(globals(), builtins());
+    plugins.unshift(remap({
+      originalPath: './test-graphql/fetch-mock-node.js',
```
minasmart: Instead of always importing something that always has the name `node`, why not import a virtual file called `isomorphic-fetch-mock`, and when you're in browser mode, point that at the browser version, when you're in node mode, point it at the node version? This means there'd be complimentary stanzas in the if and else block here, but also that the naming in the import would be more consistent.

test-graphql/client-test.js
jzjiang: Should I remove the assert.deepEqual for the queries from these tests then?

test-graphql/fetch-mock-browser.js
```diff
@@ -0,0 +1 @@
+export {default as fetchMock} from 'fetch-mock/es5/client';
```
minasmart: Why not `export default from '...'`? What was the thinking around it being an explicitly named export?

test-graphql/client-test.js
minasmart: No, that's like... future work.

test-graphql/client-test.js
```diff
@@ -42,4 +55,163 @@ suite('client-test', () => {
...
+                weight
+              }
+            }
+          }
+        }
+      }`));
+    });
+  });
 });
+
```
swalkinshaw: 🔥  remove blank line

test-graphql/fetch-mock-browser.js
jzjiang: Whoops, still getting used to the different export syntaxes. Will fix.

test-graphql/client-test.js
```diff
@@ -42,4 +55,163 @@ suite('client-test', () => {
...
+    return client.fetchAllProducts().then((products) => {
+      assert.ok(Array.isArray(products), 'products is an array');
+      assert.equal(products.length, 2, 'there are two products');
+
+      const passedArgs = fetchMock.lastCall('https://multiple-products.myshopify.com/api/graphql');
+
+      const passedQuery = JSON.parse(passedArgs[1].body).query;
+
+      assert.deepEqual(tokens(passedQuery.toString()), tokens(
```
swalkinshaw: We're thinking that asserting the generated query isn't too useful/contextual to JS Buy SDK. What's more important here is asserting things about the response like the models and properties.

It's usually better to avoid tying tests to internals of a dependency as well. graphql-js-client might change something about these generated queries.

minasmart:
@jzjiang is this PR ready for re-review?

jzjiang:
@minasmart Yes it is.

jzjiang:
This PR is ready for re-review. #297 changed most of the fetching, so the only changes on top of that are mainly schema updates.

PR #297, opened 2017-01-30
[FEATURE] Allow dynamic fields in queries

jzjiang:
Allows the user to specify which fields they want in their product/collection query.
For example,
```javascript
client.fetchCollection('369312584', collectionQuery(['title', 'updatedAt', ['image', imageQuery(['src'])]]));
```
will return the Collection with only `title`, `updatedAt` and `image` fields, where `image` contains only its `src` field.
Calling any of the fetch methods on client without supplying a query parameter will return a Product/Collection with default fields.


minasmart:
This might bring some of the internal query building up to the top level? Thoughts?

```javascript
fetchCollection('12345',
  collectionQuery(['title', 'updatedAt', 'handle'], {
    images: imageQuery(['src']),
    variants: productVariantQuery(['price', 'title'])
  })
)
```

swalkinshaw:
Whatever the API is, it should ideally be consistent.

Here's a current example:

```js
client.fetchProduct('7857989384', productQuery('id', {images: {fields: ['id', 'src']}})).then((product) => {
});
```

`fields` is not consistent in this case since it's not used in `productQuery.

@minasmart's suggestion is one way around that.

Could it be done with a plain object?

```js
client.fetchProduct('7857989384', {fields: [
  'id', {
    images: {
      fields: ['id', 'src']
    }
  }
]}).then((product) => {
});
```

minasmart:
I'd be fine with the plain object approach if every thing that was of `KIND` `Object` was the same type of thing, but some are lists, some are connections, and some are embedded objects. I have a hard time seeing a way to differentiate between them just using a hash.

jzjiang:
I just updated the code to use @minasmart's suggestion. @swalkinshaw do you have any input for plain object over this approach?

swalkinshaw:
Nope, it was just an alternative. I'm probably good with the current API you have.

I'll have a bunch of other review comments but I'll wait until we're good on the approach.

jzjiang:
@swalkinshaw brought up some good points that need to be addressed. We need to be able to add/remove arguments on top of the default without recreating the whole query. Currently, adding and removing scalars and objects to the query is inconsistent and a bit cumbersome. We've come up with a couple options below to address these issues.

### Option 1
Wrap scalars and objects in another class and create add/remove/set methods for consistency. It may be confusing for the user dealing with two explicitly separate kinds of fields. 
``` javascript
const query = productQuery();
query.scalars.add('newField');
query.objects.add('newQuery', newQuery());
query.scalars.remove('newField');
query.objects.remove('newQuery');

// Additionally query newField in images
query.objects.set('images', imageQuery().scalars.add('newField')); 
```

### Option 2
Convert each function into a class.
``` javascript
// Constructor would work in a similar manner to the current function, supplying defaults
let query = new ProductQuery();
// or taking in field selection
query = new ProductQuery(['id', 'title', 'vendor'], new ImageQuery(['src']), new OptionQuery());
query.add('field');
query.remove('field');
query.add('image', new ImageQuery()); // Optionally takes an object as a second arg
query.remove('image');
query.set('image', new ImageQuery(['src']);
```

@minasmart thoughts/other suggestions?

minasmart:
I'm leaning toward option two here. We tried the approach of splitting things into scalars and not scalars once before, and it got weird, because `objects`, or just `not scalars` could be lists, hashes, or connections, and we do some magic around connections so what we put in that part isn't really a true representation. In short, things got weird.

The one augmentation I'd apply to the query interface is placing all fields, regardless of type in a sequence (array), and when the field is more complex than a scalar, that value should be a  tuple, with the field name and function that returns the inner query.

so more like:

```javascript
let query = productQuery(['id', 'title', 'vendor', ['image', imageQuery(['src'])], 'handle']);
```

I'd also, say to start with functions over classes until we know we need mutable state.

swalkinshaw:
Would JS devs consider that tuple pattern unusual? Without any evidence my hunch would be yes.

```js
query = productQuery(['id', 'title', 'vendor', {image: imageQuery(['src'])}, 'handle']);
```

Seems like the trade off is a little more internal complexity vs easier use?

"Everything arrays" seems consistent as an API but users would need to know when to use the tuple anyway that I don't think using an object is any more overhead for them?

minasmart:
the object makes order important. You have scalars first, than non-scalars. The tuple approach let's you do anything in any order. I wanna say tuples for now, and then if it's super weird to people we can tweak the interface later. I also think it'll be the easiest to implement as a starting point.

src-graphql/client.js
```diff
@@ -12,115 +14,33 @@ export default class Client {
...
   }
 
-  fetchProduct(id) {
-    const query = this.graphQLClient.query((root) => {
-      root.add('product', {args: {id: createGid('Product', id)}}, (product) => {
-        addProductFields(product);
-      });
-    });
-
+  fetchProduct(query) {
```
minasmart: Should there be an id and default query here?

src-graphql/client.js
```diff
@@ -12,115 +14,33 @@ export default class Client {
...
   }
 
-  fetchCollection(id) {
-    const query = this.graphQLClient.query((root) => {
-      root.add('collection', {args: {id: createGid('Collection', id)}}, (collection) => {
-        addCollectionFields(collection);
-      });
-    });
-
+  fetchCollection(query) {
```
minasmart: id and default query

src-graphql/collection-query.js
```diff
@@ -0,0 +1,36 @@
+import imageQuery from './image-query';
+import createGid from './create-gid';
+
+export default function collectionQuery({client, id}, fields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]]) {
```
minasmart: maybe stick the default list in a value so it fits more cleanly on a shorter line? So like 

```javascript
const defaultFields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]];

export default function collectionQuery({client, id}, fields = defaultFields) {
```

src-graphql/collection-query.js
```diff
@@ -0,0 +1,36 @@
...
+  let query;
+
+  if (id) {
+    query = client.query((root) => {
+      root.add('collection', {args: {id: createGid('Collection', id)}}, (collection) => {
```
minasmart: there's no longer a root `collection` field. Should use the node interface and an inline fragment.

src-graphql/collection-query.js
```diff
@@ -0,0 +1,36 @@
...
+          addCollectionFields(collections, fields);
+        });
+      });
+    });
+  }
+
+  return query;
+}
+
+function addCollectionFields(collection, collectionFields) {
```
minasmart: I'm surprised the linter doesn't complain about usage before definition. The function should be declared before it's used.

src-graphql/images-query.js
```diff
@@ -0,0 +1,9 @@
+export default function imagesQuery(fields = ['id', 'src', 'altText']) {
```
minasmart: consistency in file naming and method naming. Some of these are plural and some singular.

src-graphql/product-query.js
```diff
@@ -0,0 +1,39 @@
+import optionQuery from './option-query';
+import imagesQuery from './images-query';
+import variantQuery from './variant-query';
+import createGid from './create-gid';
+
+export default function productQuery({client, id}, fields = ['id', 'createdAt', 'updatedAt', 'bodyHtml', 'handle', 'productType',
+  'title', 'vendor', 'tags', 'publishedAt', ['options', optionQuery()], ['images', imagesQuery()], ['variants', variantQuery()]]) {
+  let query;
```
minasmart: You don't need a `let` query. You can do an `if (id) { return client.query( ... ) } else { return client.query( ... ) }`

src-graphql/collection-query.js
```diff
@@ -0,0 +1,36 @@
+import imageQuery from './image-query';
+import createGid from './create-gid';
+
+export default function collectionQuery({client, id}, fields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]]) {
+  let query;
```
minasmart: You don't need this let. You can conditionally return.

src-graphql/product-query.js
```diff
@@ -0,0 +1,39 @@
...
+      });
+    });
+  }
+
+  return query;
+}
+
+function addProductFields(product, productFields) {
+  productFields.forEach((field) => {
+    if (Object.prototype.toString.call(field) === '[object String]') {
```
swalkinshaw: Maybe extract this to an `is-string` module since it's used in 3 places. We have a similar one with the graphql client for `is-object`.

swalkinshaw:
My main concern is the same as @minasmart's with `client` having to be passed in.

This PR doesn't contain the `add`, `remove`, `set` API yet. I'm fine with doing that in a separate PR though.

minasmart:
@swalkinshaw I recommended leaving out the mutability API until later. I feel  that can be an incremental addition.

src-graphql/product-query.js
```diff
@@ -0,0 +1,39 @@
+import optionQuery from './option-query';
+import imagesQuery from './images-query';
+import variantQuery from './variant-query';
+import createGid from './create-gid';
+
+export default function productQuery({client, id}, fields = ['id', 'createdAt', 'updatedAt', 'bodyHtml', 'handle', 'productType',
+  'title', 'vendor', 'tags', 'publishedAt', ['options', optionQuery()], ['images', imagesQuery()], ['variants', variantQuery()]]) {
+  let query;
+
+  if (id) {
```
swalkinshaw: This might be better split up into two separate query functions. One would takes `id` and one which doesn't.

@minasmart thoughts?

src-graphql/collection-query.js
jzjiang: Gotcha, updating schema and fixing the queries.

src-graphql/images-query.js
jzjiang: There's an imageQuery and imagesQuery because one is a connection (in products) and the other is singular (in collections). I thought about combining them within one function but that required an extra boolean to be passed with every query. Do you have any suggestions?

src-graphql/client.js
jzjiang: I had some issues with it before but currying fixes it, so adding it now.

src-graphql/product-query.js
minasmart: Yes.

src-graphql/images-query.js
minasmart: It may actually be better to be specific and use the graphql terms then. Where it's a paginated set, maybe call it an `imageConnectionQuery`, and when it's a singular image, call it `imageQuery`. This approach would mean you should rename some of your other queries too though.

jzjiang:
I think I've fixed up everything. Any other comments?

src-graphql/collection-default-fields.js
```diff
@@ -0,0 +1,3 @@
+import imageQuery from './image-query';
+
+export default ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]];
```
minasmart: Unless this is being tested separately, I don't think it needs to be its own module.

src-graphql/image-query.js
```diff
@@ -0,0 +1,9 @@
+export default function imageQuery(fields = ['id', 'src', 'altText']) {
+  return function(parentQuery, fieldName) {
+    parentQuery.add(fieldName, (image) => {
```
minasmart: This is a small thing, but the parent isn't a query, it's a `selectionSet`. `parentSelection` is probably a more accurate name.

swalkinshaw:
Agreed about the separate modules.

The rest looks good so I think this can be merged and then we can iterate on it.

PR #300, opened 2017-02-06
[FEATURE] Fetch paginated images and variants on products

jzjiang:
- `fetchAllProducts` and `fetchProduct` now fetch and return all paginated images and variants. The test fixtures for images and variants only contain 1 per page rather than 20 for simplicity. 
- This PR also includes some test cleanup on fetch-mock.


jzjiang:
Will be updated after https://github.com/Shopify/storefront-api-team-projects/issues/44

jzjiang:
This should be ready for review! 

Checking if there are more pages is a bit iffy right now because I'm not sure how we're planning on changing `hasNextPage`. I just check if the returned result is empty, which is pretty inefficient, but still works for the time being. An alternative would be to use the existing `hasNextPage` attribute, but I think we'd need to resolve the path for `response.data` like we did for `response.model` to access it without hardcoding the path.

src-graphql/client.js
```diff
@@ -22,15 +22,59 @@ export default class Client {
...
+    });
+  }
+
   fetchAllProducts(query = productConnectionQuery()) {
     return this.graphQLClient.send(query(this.graphQLClient)).then((response) => {
-      return response.model.shop.products;
+      const promises = [];
+
+      // Add all images and variants for each product
+      for (let i = 0; i < response.model.shop.products.length; i++) {
```
swalkinshaw: Think you could use a `reduce` here:

```js
const promises = response.model.shop.products.reduce((promises, product, i) => {
  if (response.data.shop.products.edges[i].node.images.pageInfo.hasNextPage) {
    promises.push(this.fetchAll(product.images));
  }

  if (response.data.shop.products.edges[i].node.variants.pageInfo.hasNextPage) {
    promises.push(this.fetchAll(product.variants));
  }

  return promises;
}, []);

return Promise.all(promises).then(() => {
  return response.model.shop.products;
});
```

src-graphql/client.js
```diff
@@ -22,15 +22,59 @@ export default class Client {
...
+      }
+
+      paginatedList.push(...result.model);
+
+      return this.fetchAll(paginatedList);
+    });
+  }
+
   fetchAllProducts(query = productConnectionQuery()) {
     return this.graphQLClient.send(query(this.graphQLClient)).then((response) => {
```
swalkinshaw: I'd probably destructure `response` into `{model,data}`

src-graphql/client.js
```diff
@@ -22,15 +22,59 @@ export default class Client {
     });
   }
 
+  fetchAll(paginatedList) {
```
swalkinshaw: maybe `paginatedModel`? Since the array is made up of `model`s

src-graphql/client.js
```diff
@@ -22,15 +22,59 @@ export default class Client {
...
+    return this.graphQLClient.fetchNextPage(paginatedList).then((result) => {
```
swalkinshaw: Destructure to `{model}`?

src-graphql/client.js
```diff
@@ -22,15 +22,59 @@ export default class Client {
...
+    return this.graphQLClient.send(query(this.graphQLClient)).then(({model, data}) => {
+      const promises = model.shop.products.reduce((promiseAcc, product, i) => {
+
+        // Fetch the rest of the images for this product
+        if (data.shop.products.edges[i].node.images && data.shop.products.edges[i].node.images.pageInfo.hasNextPage) {
+          promiseAcc.push(this.fetchAll(product.images));
+        }
+
+        // Fetch the rest of the variants for this product
+        if (data.shop.products.edges[i].node.variants && data.shop.products.edges[i].node.variants.pageInfo.hasNextPage) {
```
swalkinshaw: Can you make a const to reference `data.shop.products.edges[i].node` throughout this function?

It will just make it more readable

src-graphql/client.js
```diff
@@ -22,15 +22,61 @@ export default class Client {
     });
   }
 
+  fetchAll(paginatedModels) {
```
minasmart: I don't think this should be exposed as a public method. It's an easy path for the library user to make some very generic and long-running calls.

src-graphql/client.js
```diff
@@ -22,15 +22,61 @@ export default class Client {
...
+
   fetchAllProducts(query = productConnectionQuery()) {
-    return this.graphQLClient.send(query(this.graphQLClient)).then((response) => {
-      return response.model.shop.products;
+    return this.graphQLClient.send(query(this.graphQLClient)).then(({model, data}) => {
+      const promises = model.shop.products.reduce((promiseAcc, product, i) => {
+        const productData = data.shop.products.edges[i].node;
+
+        // Fetch the rest of the images for this product
+        if (productData.images && productData.images.pageInfo.hasNextPage) {
```
minasmart: What is the default page size for images and variants here? Just to avoid this being the common call path, can you request the max supported by the API (I think 250)?

src-graphql/client.js
```diff
@@ -22,15 +22,61 @@ export default class Client {
...
+
+      // Fetch the rest of the images for this product
+      if (productData.images && productData.images.pageInfo.hasNextPage) {
+        promises.push(this.fetchAll(response.model.node.images));
+      }
+
+      // Fetch the rest of the variants for this product
+      if (productData.variants && productData.variants.pageInfo.hasNextPage) {
+        promises.push(this.fetchAll(response.model.node.variants));
+      }
```
minasmart: Since this (L68 - L75) is a shared behaviour between the two product fetching methods, maybe extract it into a `saturateProducts` method or something?

test-graphql/client-test.js
```diff
@@ -161,12 +175,81 @@ suite('client-test', () => {
...
+      domain: 'paginated-images.myshopify.com',
+      storefrontAccessToken: 'abc123'
+    });
+
+    const client = new Client(config);
+
+    fetchMock.postOnce('https://paginated-images.myshopify.com/api/graphql', productWithPaginatedImagesFixture)
+      .postOnce('https://paginated-images.myshopify.com/api/graphql', secondPageImagesFixture)
+      .postOnce('https://paginated-images.myshopify.com/api/graphql', thirdPageImagesFixture)
+      .postOnce('https://paginated-images.myshopify.com/api/graphql', emptyPageImagesFixture);
```
minasmart: This is very thorough. 😄

src-graphql/client.js
jzjiang: It's currently 20, so I'll update to 250

jzjiang:
@minasmart I've added your suggestions. Could you take another look?

src-graphql/client.js
```diff
@@ -7,6 +7,29 @@ import productConnectionQuery from './product-connection-query';
 import collectionQuery from './collection-query';
 import collectionConnectionQuery from './collection-connection-query';
 
+function fetchAll(paginatedModels, client) {
```
minasmart: Sorry, I'd still differentiate the name here, and call it something like `fetchAllPages`, so it's clear it's not just the index, it's literally everything.

src-graphql/client.js
```diff
@@ -7,6 +7,29 @@ import productConnectionQuery from './product-connection-query';
...
+  return client.fetchNextPage(paginatedModels).then(({model}) => {
+    // Until we know how hasNextPage will be exposed, we query until the result is empty
```
minasmart: oh. I have a PR for this. Want to hang tight until it gets merged in?

src-graphql/client.js
```diff
@@ -7,6 +7,29 @@ import productConnectionQuery from './product-connection-query';
...
+      return paginatedModels;
+    }
+
+    paginatedModels.push(...model);
+
+    return fetchAll(paginatedModels, client);
+  });
+}
+
+function saturateProducts(promises, productData, product, client) {
```
minasmart: @swalkinshaw is there a better name or is this clear to you too?

src-graphql/client.js
```diff
@@ -23,14 +46,33 @@ export default class Client {
...
 
   fetchAllProducts(query = productConnectionQuery()) {
-    return this.graphQLClient.send(query(this.graphQLClient)).then((response) => {
-      return response.model.shop.products;
+    return this.graphQLClient.send(query(this.graphQLClient)).then(({model, data}) => {
+      const promises = model.shop.products.reduce((promiseAcc, product, i) => {
+        const productData = data.shop.products.edges[i].node;
+
+        // Fetch the rest of the images and variants for this product
+        saturateProducts(promiseAcc, productData, product, this.graphQLClient);
```
minasmart: could saturate return the promises, instead of mutating an input?

swalkinshaw: 👍  to this, think you could just return the `fetchAll` then `push` in here?

src-graphql/client.js
swalkinshaw: I don't mind it... but it's literally just doing some `fetchAll` calls. Something more literal would be `fetchAllProductResources` or something. If we aren't mutating the promises in here then `saturate` makes less sense I think too?

jzjiang:
I'll merge this for now and create a separate PR after Shopify/graphql-js-client#66

PR #305, opened 2017-02-20
fixed updateQuantity function

glumb:
Throwing an error when changing a variants quantity which is not a child of `product`.


minasmart:
Can you model what you're trying to prevent with a test?

rebeccajfriedman:
Closing this as the file no longer exists.

PR #307, opened 2017-02-21
[FEATURE] Query products/collections by specific fields

jzjiang:
This PR adds `fetchQueryProducts` and `fetchQueryCollections` to loosely match the [REST versions](https://shopify.github.io/js-buy-sdk/api/classes/ShopClient.html#method-fetchQueryCollections). This acts as a layer over the `query` arg on products and collections so users can search by specific fields, like `title` and `updatedAt`.

### DIFF
--- a/test/client-test.js
+++ b/test/client-test.js
@@ -382,6 +382,26 @@ suite('client-test', () => {
     fetchMock.postOnce('https://sorted-query.myshopify.com/api/graphql', shopWithSortedProductsFixture);
 
     return client.fetchQueryProducts({sortBy: 'updatedAt', sortDirection: 'desc'}, productConnectionQuery(['updatedAt'])).then((products) => {
+      const [_arg, {body}] = fetchMock.lastCall('https://sorted-query.myshopify.com/api/graphql');
+      const queryString = `query {
+        shop {
+          products (first: 20, sortKey: UPDATED_AT, reverse: true, query: "") {
+            pageInfo {
+              hasNextPage
+              hasPreviousPage
+            }
+            edges {
+              cursor
+              node {
+                id
+                updatedAt
+              }
+            }
+          }
+        }
+      }`;
+
+      assert.deepEqual(tokens(JSON.parse(body).query), tokens(queryString));
       assert.equal(products.length, 3);
       assert.equal(products[0].updatedAt, shopWithSortedProductsFixture.data.shop.products.edges[0].node.updatedAt);
       assert.equal(products[1].updatedAt, shopWithSortedProductsFixture.data.shop.products.edges[1].node.updatedAt);
@@ -400,6 +420,26 @@ suite('client-test', () => {
     fetchMock.postOnce('https://sorted-query.myshopify.com/api/graphql', shopWithSortedCollectionsFixture);
 
     return client.fetchQueryCollections({sortBy: 'title'}, collectionConnectionQuery(['title'])).then((collections) => {
+      const [_arg, {body}] = fetchMock.lastCall('https://sorted-query.myshopify.com/api/graphql');
+      const queryString = `query {
+        shop {
+          collections (first: 20, sortKey: TITLE, query: "") {
+            pageInfo {
+              hasNextPage
+              hasPreviousPage
+            }
+            edges {
+              cursor
+              node {
+                id
+                title
+              }
+            }
+          }
+        }
+      }`;
+
+      assert.deepEqual(tokens(JSON.parse(body).query), tokens(queryString));
       assert.equal(collections.length, 3);
       assert.equal(collections[0].title, shopWithSortedCollectionsFixture.data.shop.collections.edges[0].node.title);
       assert.equal(collections[1].title, shopWithSortedCollectionsFixture.data.shop.collections.edges[1].node.title);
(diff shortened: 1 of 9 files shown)


src-graphql/client.js
```diff
@@ -87,4 +87,67 @@ export default class Client {
       return response.model.node;
     });
   }
+
+  fetchQueryProducts(queryObject, query = productConnectionQuery()) {
```
swalkinshaw: Default here? `queryObject = {}`. Should eliminate the need for the `queryObject` conditional.

src-graphql/client.js
```diff
@@ -87,4 +87,67 @@ export default class Client {
...
+      if (queryObject.updatedAtMin) {
+        queryArgString = queryArgString.concat(`updated_at:>='${queryObject.updatedAtMin}' `);
+      }
+      if (queryObject.createdAtMin) {
+        queryArgString = queryArgString.concat(`created_at:>='${queryObject.createdAtMin}' `);
+      }
+      if (queryObject.productType) {
+        queryArgString = queryArgString.concat(`product_type:'${queryObject.productType}' `);
+      }
+      if (queryObject.limit) {
```
swalkinshaw: @minasmart should we keep the `limit` name or just use `first` to match up with GraphQL?

src-graphql/collection-connection-query.js
```diff
@@ -4,10 +4,10 @@ import addFields from './add-fields';
 const defaultFields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]];
 
 export default function collectionConnectionQuery(fields = defaultFields) {
-  return function(client) {
+  return function(client, options) {
     return client.query((root) => {
       root.add('shop', (shop) => {
-        shop.addConnection('collections', {args: {first: 20}}, (collections) => {
+        shop.addConnection('collections', {args: Object.assign({}, {first: 20}, options)}, (collections) => {
```
swalkinshaw: The first empty object can be skipped? `Object.assign({first: 20}, options)`

src-graphql/product-connection-query.js
```diff
@@ -7,10 +7,10 @@ const defaultFields = ['id', 'createdAt', 'updatedAt', 'bodyHtml', 'handle', 'pr
   'publishedAt', ['options', optionQuery()], ['images', imageConnectionQuery()], ['variants', variantConnectionQuery()]];
 
 export default function productConnectionQuery(fields = defaultFields) {
-  return function(client) {
+  return function(client, options) {
     return client.query((root) => {
       root.add('shop', (shop) => {
-        shop.addConnection('products', {args: {first: 20}}, (products) => {
+        shop.addConnection('products', {args: Object.assign({}, {first: 20}, options)}, (products) => {
```
swalkinshaw: Same here

test-graphql/client-test.js
```diff
@@ -252,4 +254,38 @@ suite('client-test', () => {
...
+    const config = new Config({
+      domain: 'query-products.myshopify.com',
+      storefrontAccessToken: 'abc123'
+    });
+
+    const client = new Client(config);
+
+    fetchMock.postOnce('https://query-products.myshopify.com/api/graphql', {data: {shop: {products: {edges: [{node: {title: 'Cat'}}]}}}});
+
+    return client.fetchQueryProducts({title: 'Cat', limit: 10}, productConnectionQuery(['title'])).then((products) => {
```
swalkinshaw: Can these tests have every option set? It's more comprehensive that way

jzjiang: Do you think it'd also be useful to check the query sent to see if the query arg string is being built properly?

swalkinshaw: Yeah sorry that's what I was getting at :)

swalkinshaw:
This PR itself looks good, but we might have to rethink the feature.

I've realized that the current query params supported by JS Buy SDK are just the ones our REST API supports. They can be found here: https://help.shopify.com/api/reference/product#index

We could just map all these the way you're doing now, but I wonder if we should wait on this a little bit to see if we can improve the schema.

Or we can just keep this as is for now and update later.

@minasmart thoughts?

minasmart:
I'd keep this as is, but with the knowledge that we're gonna rework things. I don't like how we're doing this in our schema, but until that gets changed, this is ok.

src-graphql/client.js
```diff
@@ -85,4 +85,115 @@ export default class Client {
...
+      queryArgStrings.push(`product_type:'${queryObject.productType}'`);
+    }
+    if (queryObject.limit) {
+      options.first = queryObject.limit;
+    }
+    if (queryObject.sortBy) {
+      let sortKey;
+
+      switch (queryObject.sortBy) {
+        case 'title-ascending':
```
swalkinshaw: Wondering if we should consider changing how consumers specify sort keys and order. Right now it's a little strange to have a single string that contains both concepts.

Thinking of something like:

```js
client.fetchQueryProducts({sortBy: 'updatedAt', sortDirection: 'desc'})
```

This matches up with the key more directly.

@minasmart thoughts? If we want to change it, I'm indifferent to making it part of this PR or in a future one.

jzjiang:
Rebased this PR and updated with @swalkinshaw's comment. Could I get a quick re-review?

test/client-test.js
```diff
@@ -273,6 +285,128 @@ suite('client-test', () => {
...
+
+      assert.deepEqual(tokens(JSON.parse(body).query), tokens(queryString));
+      assert.equal(collections.length, 1);
+      assert.equal(collections[0].title, queryCollectionFixture.data.shop.collections.edges[0].node.title);
+      assert.equal(collections[0].updatedAt, queryCollectionFixture.data.shop.collections.edges[0].node.updatedAt);
+      assert.ok(fetchMock.done());
+    });
+  });
+
+  test('it can return queried products sorted', () => {
```
swalkinshaw: Shouldn't this test and the below be asserting the query string?

PR #314, opened 2017-03-09
[FIX] Paginate using hasNextPage

jzjiang:
Update to #300 - we use the `hasNextPage` property on each node to paginate instead of querying until an empty list.


minasmart:
looks like this is missing a `yarn.lock` update, otherwise 👍

PR #316, opened 2017-03-14
[FIX] Don't paginate on empty sets

jzjiang:
If images or variants were empty, this would throw an error. Added a check to make sure we don't try to paginate on empty sets.

### DIFF
--- a/src-graphql/client.js
+++ b/src-graphql/client.js
@@ -24,11 +24,11 @@ function fetchAllProductResources(product, client) {
   const images = product.images;
   const variants = product.variants;
 
-  if (images && images[images.length - 1].hasNextPage.valueOf()) {
+  if (images && images.length && images[images.length - 1].hasNextPage.valueOf()) {
     promises.push(fetchAllPages(product.images, client));
   }
 
-  if (variants && variants[variants.length - 1].hasNextPage.valueOf()) {
+  if (variants && variants.length && variants[variants.length - 1].hasNextPage.valueOf()) {
     promises.push(fetchAllPages(product.variants, client));
   }
 
--- a/test-graphql/client-test.js
+++ b/test-graphql/client-test.js
@@ -250,4 +250,20 @@ suite('client-test', () => {
       assert.ok(fetchMock.done());
     });
   });
+
+  test('it does not fetch paginated images if the images query result was empty', () => {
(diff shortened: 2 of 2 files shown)


src-graphql/client.js
```diff
@@ -24,11 +24,11 @@ function fetchAllProductResources(product, client) {
   const images = product.images;
   const variants = product.variants;
 
-  if (images && images[images.length - 1].hasNextPage.valueOf()) {
+  if (images && images.length && images[images.length - 1].hasNextPage.valueOf()) {
```
swalkinshaw: Unrelated, but is `valueOf` needed for these? I thought `hasNextPage` is just a boolean?

edit:

```js
function hasNextPage(connection, edge) {
  if (edge !== connection.edges[connection.edges.length - 1]) {
    return true;
  }

  return connection.pageInfo.hasNextPage;
}
```

Maybe `valueOf` should be called in this function on `connection.pageInfo.hasNextPage` to be consistent.

jzjiang: Scalars aren't being unwrapped properly by `Boolean` for some reason. I think @minasmart was looking into this.

minasmart: Yeah, truthy and falsey tests don't actually attempt to coerce the value to a boolean. I'm reading up on it, but boxed primitives work mostly as you'd expect until you deal with booleans. I'm trying to think of other ways to do this without boxing everything. If you have any suggestions around this, please let me know.

swalkinshaw: Yeah I tried that myself and it didn't work either :(

In this case I just mean that a function like `hasNextPage` shouldn't return a primitive `Boolean` *or* a `Scalar`. So I think the solution here is for the client to call `connection.pageInfo.hasNextPage.valueOf()`. Then the return type is consistent.

minasmart: isn't calling `connection.pageInfo.hasNextPage.valueOf()` what's going on?

minasmart: or do you mean to attach page info back to the set, and not load it onto every node?

swalkinshaw: I'm referring to graphql-js-client where the `hasNextPage` function is. The return type should be consistent there so consumers like JS Buy SDK shouldn't have to worry about it.

PR #317, opened 2017-03-14
[FEATURE] Add Product helper to return variant based on options

jzjiang:
This PR adds a Product helper to return the variant of a product based on the options selected. 
Usage:
```js
client.Product.Helpers.variantForOptions(product, {Fur: 'Fluffy', Size: 'Medium'});
```

I'll be adding `client.Checkout.Mutation` functions in a later PR following a similar structure.

### DIFF
--- a/src-graphql/product-helpers.js
+++ b/src-graphql/product-helpers.js
@@ -1,17 +1,9 @@
 export default class ProductHelpers {
   variantForOptions(product, options) {
-    const [selectedVariant] = product.variants.reduce((variantAcc, variant) => {
-      const match = variant.selectedOptions.every((selectedOption) => {
+    return product.variants.find((variant) => {
+      return variant.selectedOptions.every((selectedOption) => {
         return options[selectedOption.name] === selectedOption.value.valueOf();
       });
-
-      if (match) {
-        variantAcc.push(variant);
-      }
-
-      return variantAcc;
-    }, []);
-
-    return selectedVariant;
+    });
   }
 }
--- a/src-graphql/product-helpers.js
+++ b/src-graphql/product-helpers.js
@@ -1,0 +1,17 @@
+export default class ProductHelpers {
+  variantForOptions(product, options) {
+    const [selectedVariant] = product.variants.reduce((variantAcc, variant) => {
+      const match = variant.selectedOptions.every((selectedOption) => {
+        return options[selectedOption.name] === selectedOption.value.valueOf();
+      });
+
+      if (match) {
+        variantAcc.push(variant);
+      }
+
+      return variantAcc;
+    }, []);
+
+    return selectedVariant;
+  }
+}
--- a/test-graphql/product-helpers-test.js
+++ b/test-graphql/product-helpers-test.js
@@ -1,0 +1,27 @@
+import assert from 'assert';
+import Config from '../src-graphql/config';
+import Client from '../src-graphql/client';
+import ProductHelpers from '../src-graphql/product-helpers';
+import singleProductFixture from '../fixtures/product-fixture';
+import fetchMock from './isomorphic-fetch-mock'; // eslint-disable-line import/no-unresolved
+
+suite('product-helpers-test', () => {
+  const productHelpers = new ProductHelpers();
+
+  test('it returns the variant based on options given', () => {
+    const config = new Config({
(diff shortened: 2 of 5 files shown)


swalkinshaw:
Is this purely for consumers of the library? or are we going to use it internally anywhere?

jzjiang:
I think only for consumers of the library.

minasmart:
This is used for when people build a UI. It's mean to make our users map options back to variants.

src-graphql/product-helpers.js
```diff
@@ -0,0 +1,17 @@
+export default class ProductHelpers {
+  variantForOptions(product, options) {
+    const [selectedVariant] = product.variants.reduce((variantAcc, variant) => {
```
minasmart: couldn't this just be `.find`? Then just `return product.variants.find(...)`

test-graphql/product-helpers-test.js
```diff
@@ -0,0 +1,27 @@
...
+    const client = new Client(config);
+
+    fetchMock.postOnce('https://single-product.myshopify.com/api/graphql', singleProductFixture);
+
+    return client.fetchProduct('7857989384').then((product) => {
+      const variant = productHelpers.variantForOptions(product, {Fur: 'Fluffy', Size: 'Medium'});
+
+      assert.equal(variant.id, 'gid://shopify/ProductVariant/25602235976');
+    });
+  });
```
minasmart: You should add a test case for where the variant doesn't exist (not all combinations of options are valid. You can manually remove a particular variant)

src-graphql/product-helpers.js
jzjiang: yep

src-graphql/product-helpers.js
```diff
@@ -0,0 +1,11 @@
+export default class ProductHelpers {
+  variantForOptions(product, options) {
+    const selectedVariant = product.variants.find((variant) => {
```
swalkinshaw: Just `return` this directly?

test-graphql/product-helpers-test.js
```diff
@@ -0,0 +1,38 @@
...
+  const productHelpers = new ProductHelpers();
+  const config = new Config({
+    domain: 'single-product.myshopify.com',
+    storefrontAccessToken: 'abc123'
+  });
+
+  const client = new Client(config);
+
+  test('it returns the variant based on options given', () => {
+    fetchMock.postOnce('https://single-product.myshopify.com/api/graphql', singleProductFixture);
```
swalkinshaw: As a separate improvement, we should be able to use a fixture plus it's query and decode it to get the `product` graph model result *without* having to go through a client and do fetch mocking. Ideally in a unit test we can avoid all of that.

Something like:

```js
const product = decode(query, singleProductFixture); // or even a fixture to import which has already done this
const variant = productHelpers.variantForOptions(product, {Fur: 'Fluffy', Size: 'Medium'});

assert.equal(variant.id, 'gid://shopify/ProductVariant/25602235976');
```

jzjiang: Yeah I don't like how it's generating the model either. I can create a separate PR in `graphql-js-client` to export `decode` so I can properly generate the models.

jzjiang:
@minasmart any other comments?

test-graphql/product-helpers-test.js
minasmart: attaching decode to the root graphql client is probably an ok thing to do 😄

test-graphql/product-helpers-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+    assert.equal(variant.id, 'gid://shopify/ProductVariant/25602235976');
+    assert.ok(fetchMock.done());
+  });
+
+  test('it returns undefined if the variant does not exist', () => {
+    const variant = productHelpers.variantForOptions(productModel.node, {Fur: 'Fluffy', Size: 'Small'});
+
+    assert.equal(typeof variant, 'undefined');
+    assert.ok(fetchMock.done());
```
swalkinshaw: These can be removed? And the import?

jzjiang: whoops, yes you are right. thanks!

jzjiang:
@minasmart I've updated the tests. Everything look good to you?

PR #321, opened 2017-03-20
[FEATURE] Checkout Create Mutation

jzjiang:
Main method added is `createCheckout()` in `client.js`. Naming could be changed later (i.e. `client.Checkout.Mutations.create()`), but it is currently a top level function for consistency with the other fetch methods.

This PR also:
- Moves some of the query building into `base-query.js` that other queries call
- Adds query functions for checkout
- Updates the schema

### Example Usage
```js
const input = {
   lineItems: [
     {variantId: 'gid://shopify/ProductVariant/1', quantity: 5, customAttributes: [{key: 'hi', value: 'bye'}]}
    ],
    shippingAddress: {
     address1: '123 Cat Road',
     city: 'Cat Land',
     company: 'Catmart',
     country: 'Canada',
     firstName: 'Meow',
     lastName: 'Meowington',
     phone: '4161234566',
     province: 'ON',
     zip: 'M3O 0W1'
  }
};

client.createCheckout(input);
// or specify fields on the checkout query
client.createCheckout(input, checkoutQuery(['id', 'createdAt']));
```

Github issue: Shopify/storefront-api-team-projects#60


src-graphql/checkout-query.js
```diff
@@ -0,0 +1,11 @@
+import lineItemConnectionQuery from './line-item-connection-query';
+import shippingRateQuery from './shipping-rate-query';
+import mailingAddressQuery from './mailing-address-query';
+import baseQuery from './base-query';
+
+const defaultFields = ['id', 'ready', 'note', 'createdAt', 'updatedAt', 'requiresShipping', ['shippingLine', shippingRateQuery()],
```
swalkinshaw: Can we format this in the long form?

```js
const defaultFields = [
  'id',
  'ready',
  // etc
];
```

Easier to read and modify in the future

src-graphql/client.js
```diff
@@ -109,4 +102,50 @@ export default class Client {
...
+        query(checkoutCreate, 'checkout');
+      });
+    });
+
+    return this.graphQLClient.send(mutation).then((result) => {
+      // Paginate on line items
+      const promises = [];
+      const lineItems = result.model.checkoutCreate.checkout.lineItems;
+
+      if (lineItems && lineItems.length && lineItems[lineItems.length - 1].hasNextPage.valueOf()) {
```
swalkinshaw: This is probably a sign we need another function to improve this use case. Might even make sense to have it in graphql-js-client.

Don't worry about it for this PR though.

src-graphql/mailing-address-query.js
```diff
@@ -0,0 +1,7 @@
+import baseQuery from './base-query';
+
+const defaultFields = ['address1', 'address2', 'city', 'company', 'country', 'firstName', 'formatted', 'lastName', 'latitude', 'longitude', 'phone', 'province', 'zip', 'name', 'countryCode', 'provinceCode', 'id'];
```
swalkinshaw: Long form for these too

src-graphql/variant-query.js
```diff
@@ -0,0 +1,9 @@
+import selectedOptionQuery from './selected-option-query';
+import imageQuery from './image-query';
+import baseQuery from './base-query';
+
+const defaultFields = ['id', 'title', 'price', 'weight', ['image', imageQuery()], ['selectedOptions', selectedOptionQuery()]];
```
swalkinshaw: Here too please 👼

src-graphql/attribute-query.js
```diff
@@ -0,0 +1,5 @@
+import baseQuery from './base-query';
+
+export default function attributeQuery(fields = ['key', 'value']) {
```
minasmart: suggesting a rename to `lineItemAttributeQuery`, since `attributeQuery` is pretty ambiguous.

src-graphql/client.js
minasmart: with the new version of the client `valueOf` should no longer be necessary.

However, we should definitely expose a fetchAllPages method with a page size param in `graphql-js-client`

src-graphql/attribute-query.js
jzjiang: It can be used to query on checkouts as well (which I actually forgot to add and will add now). Maybe `customAttributeQuery` since that's what the field name is called on the [schema](https://github.com/Shopify/shopify/blob/master/db/graphql/storefront_schema.graphql#L28)?

minasmart: yep. if that's what's on the schema, that's probably good.

src/client.js
```diff
@@ -7,35 +7,28 @@ import productConnectionQuery from './product-connection-query';
...
-
-    if (!model[model.length - 1].hasNextPage.valueOf()) {
-      return paginatedModels;
-    }
-
-    return fetchAllPages(paginatedModels, client);
-  });
-}
+import checkoutQuery from './checkout-query';
+import fetchAllPages from './fetch-all-pages';
```
minasmart: Shouldn't this rely on the feature you pushed up stream into the graphql client?

jzjiang:
@minasmart Could you take a final look at this?

src/client.js
```diff
@@ -109,4 +100,47 @@ export default class Client {
...
+          userErrors.add('message');
+          userErrors.add('field');
+        });
+        query(checkoutCreate, 'checkout');
+      });
+    });
+
+    return this.graphQLClient.send(mutation).then((result) => {
+      // Paginate on line items
+      const promises = [this.graphQLClient.fetchAllPages(result.model.checkoutCreate.checkout.lineItems, {pageSize: 250}).then((lineItems) => {
```
minasmart: does this need to be `promises` now? 

could you not just

```javascript
return this.graphQLClient.send(mutation).then(({model}) => {
  return this.graphQLClient.fetchAllPages(model.checkoutCreate.checkout.lineItems, {pageSize: 250}).then((lineItems) => {
    model.checkoutCreate.checkout.attrs.lineItems = lineItems;

    return model.checkoutCreate.checkout;
  });
});
```

swalkinshaw:
> Also, something for a separate PR, should we reject the promise if userErrors is non-null?

At first thought, this seems like a good idea. It's consistent with a RESTful API returning a non-200 code for a request with errors.

PR #324, opened 2017-03-23
Move things into place, and get ready to start releasing a beta (part 1)

minasmart:


minasmart:
@swalkinshaw and @jzjiang could you take a look here? Mostly deletions, but working toward a releasable thing.

PR #326, opened 2017-03-28
[FEATURE] Add line item mutation

jzjiang:
- Added `addLineItems` method. 
- Updated schema and added the extra fields to checkout.


minasmart:
Test failure looks like a missing null check in graphql js client where an instance of a node may be null.

PR #328, opened 2017-03-30
Change classification to Library

RyanBrushett:
Since this is a library and won't have an actual runtime (e.g. it doesn't need an oncall rotation, etc.), it is more accurately classified as such!

@minasmart @swalkinshaw 

cc @Shopify/operational-excellence 

<3


jarthorn:
@minasmart yes it was undocumented 😿   Fix: https://github.com/Shopify/services-db/pull/957

PR #329, opened 2017-03-30
[FEATURE] Add helper to resize images

jzjiang:
Adds a helper to resize images (by appending the max dimensions to the url). Unfortunately it won't work on an image that was already resized, and I'm a bit iffy on adding that in case the image they use have `some-image-name_500x500.jpg` in the original title.

### DIFF
--- a/src/image-helpers.js
+++ b/src/image-helpers.js
@@ -1,0 +1,29 @@
+/**
+ * @class ImageHelpers
+ */
+export default class ImageHelpers {
+
+  /**
+   * Generates the image src for a resized image with maximum dimensions `maxWidth` and `maxHeight`.
+   * Images do not scale up.
+   *
+   * @method imageForSize
+   * @param {Object} image The original image model to generate the image src for
+   * @param {Object} options An options object containing:
+   *  @param {Integer} options.maxHeight The maximum height for the image
+   *  @param {Integer} options.maxWidth The maximum width for the image
+   * @return {String} The image src for the resized image
+   */
+  imageForSize(image, {maxHeight, maxWidth}) {
+    const urlTokens = image.src.split('?');
+
+    // Use the section before the query
+    const imageTokens = urlTokens[0].split('.');
+
+    // Take the second last token and append the dimensions
+    imageTokens[imageTokens.length - 2] = imageTokens[imageTokens.length - 2].concat(`_${maxHeight}x${maxWidth}`);
+    urlTokens[0] = imageTokens.join('.');
+
+    return urlTokens.join('?');
+  }
+}
--- a/src/image-helpers.js
+++ b/src/image-helpers.js
@@ -15,15 +15,16 @@ export default class ImageHelpers {
    * @return {String} The image src for the resized image
    */
   imageForSize(image, {maxHeight, maxWidth}) {
-    const urlTokens = image.src.split('?');
+    const [notQuery, query] = image.src.split('?');
 
     // Use the section before the query
-    const imageTokens = urlTokens[0].split('.');
+    const imageTokens = notQuery.split('.');
 
-    // Take the second last token and append the dimensions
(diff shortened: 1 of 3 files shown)


src/image-helpers.js
```diff
@@ -0,0 +1,24 @@
...
+   * @param {Object} image the original image object to generate the image src for
+   * @param {Integer} maxHeight the maximum height for the image
+   * @param {Integer} maxWidth the maximum width for the image
+   * @return {String} The image src for the resized image
+   */
+  imageForSize(image, maxHeight, maxWidth) {
+    const tokens = image.src.split('.');
+
+    // Take the second last token and append the dimensions
+    tokens[tokens.length - 2] = tokens[tokens.length - 2].concat(`_${maxHeight}x${maxWidth}`);
```
minasmart: I think it may be safer to split on `/` (sorry). `/` has specific meaning in the URL, and likely won't be added arbitrarily. `.` could easily be added in URL params for concerns we're not aware of.

src/image-helpers.js
```diff
@@ -0,0 +1,24 @@
...
+ * @class ImageHelpers
+ */
+export default class ImageHelpers {
+
+  /**
+   * Generates the image src for a resized image with maximum dimensions `maxWidth` and `maxHeight`.
+   * Images do not scale up.
+   *
+   * @method imageForSize
+   * @param {Object} image the original image object to generate the image src for
```
minasmart: `the image model`

src/image-helpers.js
```diff
@@ -0,0 +1,26 @@
...
+   * Generates the image src for a resized image with maximum dimensions `maxWidth` and `maxHeight`.
+   * Images do not scale up.
+   *
+   * @method imageForSize
+   * @param {Object} image the original image model to generate the image src for
+   * @param {Integer} maxHeight the maximum height for the image
+   * @param {Integer} maxWidth the maximum width for the image
+   * @return {String} The image src for the resized image
+   */
+  imageForSize(image, maxHeight, maxWidth) {
```
swalkinshaw: Think I'd prefer option args here. Especially because it will be easy to confused the order of height and width.

src/image-helpers.js
```diff
@@ -0,0 +1,26 @@
...
+    const urlTokens = image.src.split('/');
+    const imageTokens = urlTokens[urlTokens.length - 1].split('.');
+
+    // Take the first token and append the dimensions
+    imageTokens[0] = imageTokens[0].concat(`_${maxHeight}x${maxWidth}`);
```
swalkinshaw: Does this assumption hold if there's multiple dots?

`cat.foo.bar.jpg`. Should it still append the dimensions on the first dot?

jzjiang: It should probably go before the last dot in that case, but I'm not sure if that holds for everything either and would be the same as just splitting on '.' like I had before.

jzjiang:
Updated to split on `?` then `.` so we don't accidentally append things to the query params.

src/image-helpers.js
```diff
@@ -0,0 +1,29 @@
...
+   * @return {String} The image src for the resized image
+   */
+  imageForSize(image, {maxHeight, maxWidth}) {
+    const urlTokens = image.src.split('?');
+
+    // Use the section before the query
+    const imageTokens = urlTokens[0].split('.');
+
+    // Take the second last token and append the dimensions
+    imageTokens[imageTokens.length - 2] = imageTokens[imageTokens.length - 2].concat(`_${maxHeight}x${maxWidth}`);
```
minasmart: pull this off as a var instead of doing the lookup twice.

jzjiang: I'd still end up doing the lookup twice because I'd only be able to pull the right side lookup into a var since I need to assign into the array. Should I still pull it into a var just to make it a bit more readable?

PR #333, opened 2017-04-04
[FEATURE] v1 alpha build, in all versions.

minasmart:
This is an isomorphic build with zero polyfills. `main` points at `index.js` (which is commonjs). `jsnext:main` points at `index.es.js`

Additionally available builds can be imported as `shopify-buy/index.amd` and `shopify-buy/index.umd`.

__NOTE:__ This will be published as a tagged release so like `npm publish --tag alpha`, which means anyone with the tag `shopify-buy@latest` will not get this. In examples, you should be able to reference this as `shopify-buy@alpha`


minasmart:
also, this will be published as a tagged release so like `npm publish --tag alpha`, which means anyone with the tag `shopify-buy@latest` will not get this. In examples, you should be able to reference this as `shopify-buy@alpha`

profiled-types.json
```diff
@@ -0,0 +1,46 @@
...
+    "ProductVariantEdge",
+    "QueryRoot",
+    "SelectedOption",
+    "ShippingRate",
+    "Shop",
+    "String",
+    "URL",
+    "UserError"
+  ]
+}
```
minasmart: This is a generated file. New lines are gonna be weird.

test/setup.js
```diff
@@ -0,0 +1,30 @@
+import GraphQLJSClient from '../src/graphql-client';
+
+function recordTypes() {
+  const types = GraphQLJSClient.trackedTypes();
+
+  if (typeof require === 'function') {
+    const body = JSON.stringify({'profiled-types': types}, null, '  ');
```
swalkinshaw: Don't care either way, but I'm just used to seeing a number for the last arg here: `2`

package.json
```diff
@@ -1,37 +1,41 @@
...
     "print-start-message": "wait-on file:.tmp/test/index.html file:.tmp/test/tests.js tcp:35729 tcp:4200 file:types.js && echo \"\n\n⚡️⚡️⚡️  Good to go at http://localhost:4200 ⚡️⚡️⚡️\"",
     "schema:refresh": "rimraf types.js && yarn run schema:fetch && yarn run schema:transform-to-types",
     "schema:fetch": "graphql-js-schema-fetch --url 'https://graphql.myshopify.com/api/graphql' --header 'Authorization: Basic MzUxYzEyMjAxN2QwZjJhOTU3ZDMyYWU3MjhhZDc0OWM=' | jq '.' > schema.json",
-    "schema:transform-to-types": "graphql-js-schema --schema-file schema.json --outdir . --schema-bundle-name 'Types' --bundle-only"
+    "schema:transform": "concurrently 'yarn run schema:transform-to-types' 'yarn run schema:transform-to-optimized-types'",
+    "schema:transform-to-types": "graphql-js-schema --schema-file schema.json --outdir . --schema-bundle-name 'Types' --bundle-only",
+    "schema:transform-to-optimized-types": "graphql-js-schema $(node scripts/profiled-types-to-whitelist-args.js) --schema-file schema.json --outdir . --schema-bundle-name 'OptimizedTypes' --bundle-only"
```
swalkinshaw: An alternate to this is skipping both JSON generation and `scripts/profiled-types-to-whitelist-args.js`.

You could dump the list as a comma separated file and use: 
```bash
graphql-js-schema --whitelist-type={$(cat profiled-types.csv)}
```
(most shells expand that pattern into calling the arg multiple times)

test/setup.js
minasmart: I thought it had to be the character literal. I'll use a number, since that's more readable than arbitrary spaces.

package.json
minasmart: `yarn run` doesn't do pattern matching, and this changes the argument format, so it has to be a set of types with no spaces and only commas. Way more error prone.

minasmart: also, it's a json file, as opposed to a CSV, since we've already proposed profiling type fields as well as types. the JSON is more extensible.

swalkinshaw: Minimist supports it transparently. This worked without any changes:

```bash
"schema:test": "graphql-js-schema --whitelist-type={$(cat types.csv)} --schema-file schema.json --outdir . --schema-bundle-name 'OptimizedTypes' --bundle-only"
```

I'm good with the argument about wanting JSON for other reasons though if it will be used for other purposes 👍

minasmart: 👍

jzjiang:
Works with ember examples and node example! 👍

PR #335, opened 2017-04-05
[FEATURE] Add method to fetch shop information

jzjiang:
Adds a method to fetch the fields on shop. Also changes `before` and `after` to `setup` and `teardown`, so mocha doesn't error out when running tests in browser.

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -12,19 +12,21 @@ function recordTypes() {
   }
 }
 
-setup(() => {
-  if (!GraphQLJSClient.startTracking) {
-    return;
-  }
+if (typeof before === 'function' && typeof after === 'function') {
+  before(() => {
+    if (!GraphQLJSClient.startTracking) {
+      return;
+    }
 
-  GraphQLJSClient.startTracking();
-});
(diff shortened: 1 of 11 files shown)


test/setup.js
```diff
@@ -12,15 +12,15 @@ function recordTypes() {
...
+setup(() => {
   if (!GraphQLJSClient.startTracking) {
     return;
   }
 
   GraphQLJSClient.startTracking();
 });
 
-after(() => {
+teardown(() => {
```
minasmart: these aren't equivalent. If anything, skip them in the browser. `before` and `after` run before the test suit and after the test suite. `setup` and `teardown` run before _every_ test and after _every_ test. Maybe just wrap these up in an `if (before && after)`

swalkinshaw:
This doesn't exist on the current version right? Just wondering about use case.

Not that it should stop us from adding features, but I realized this would obviously increase the build size both in terms of code here and the type bundle.

jzjiang:
It doesn't exist on the current version, but we'd like to be able to display information like the shop's name on the examples (e.g. `${shopName}: Ember Example`).

PR #337, opened 2017-04-10
Use shipit to publish alpha npm releases

jzjiang:


scripts/deploy
```diff
@@ -0,0 +1,77 @@
...
+
+Uploader.prototype.s3PathForFile = function (localPath, tag) {
+  return path.join(S3_DIR, tag, localPath);
+};
+
+Uploader.prototype.contentType = function(path) {
+  return mime.lookup(path) || DEFAULT_CONTENT_TYPE;
+};
+
+new Uploader().uploadDistAssets();
```
minasmart: Let's not deploy to CDN just yet.

PR #338, opened 2017-04-10
Switch from jsnext:main to module

swalkinshaw:
`module` supersedes `jsnext:main`.

See https://github.com/rollup/rollup/wiki/pkg.module

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0-alpha.2",
   "description": "The JS Buy SDK is a lightweight library that allows you to build ecommerce into any website. It is based on Shopify's API and provides the ability to retrieve products and collections from your shop, add products to a cart, and checkout.",
   "main": "index.js",
-  "jsnext:main": "index.es.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0-alpha.2",
   "description": "The JS Buy SDK is a lightweight library that allows you to build ecommerce into any website. It is based on Shopify's API and provides the ability to retrieve products and collections from your shop, add products to a cart, and checkout.",
   "main": "index.js",
-  "jsnext:main": "index.es.js",
+  "module": "index.es.js",
```
minasmart: maybe backport this to the other rollup-ed modules as well? `graphql-js-client`, and `graphql-js-schema`?

minasmart: also, maybe switch our instances of `nodeResolve` to use `module: true` instead of `jsnext: true`

swalkinshaw: I've updated two other places with it 👍  I'll do the other libs as well

PR #339, opened 2017-04-10
Use X-Shopify-Storefront-Access-Token header

swalkinshaw:
We want to standardize on our custom header instead of basic auth.


swalkinshaw:
Note: do not merge yet

src/base64encode.js
```diff
@@ -1,9 +0,0 @@
-export default function base64Encode(string) {
-  if (typeof btoa === 'function') {
-    return btoa(string);
-  } else if (typeof Buffer === 'function') {
-    return Buffer.from(string).toString('base64');
-  } else {
-    throw new Error('No native way to base64 encode');
-  }
-}
```
minasmart: removing the need for this removes the need for a polyfill. Could you make sure `Base64` isn't in the package.json?

swalkinshaw: Good catch, removed

swalkinshaw:
Support for this header has shipped 🚀

minasmart:
🚢  and tag as `alpha.3`

PR #340, opened 2017-04-10
[FEATURE] Add fetchCheckout

jzjiang:
This PR adds a `fetchCheckout` method to the client to fetch checkouts by ID.  Most changes are renaming `productQuery` --> `productNodeQuery` and `collectionQuery` --> `collectionNodeQuery` to more accurately reflect the queries.


src/checkout-node-query.js
```diff
@@ -0,0 +1,41 @@
...
+  'taxExempt',
+  'taxesIncluded',
+  'currencyCode',
+  'totalTax',
+  'subtotalPrice',
+  'totalPrice',
+  'completedAt',
+  'createdAt',
+  'updatedAt'
+];
```
swalkinshaw: Is this every field on a checkout? Should we include them all by default?

src/checkout-node-query.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default function checkoutNodeQuery(fields = defaultFields) {
+  return function(parentSelection, fieldName, id) {
+    parentSelection.add(fieldName, {args: {id: createGid('Checkout', id)}}, (node) => {
+      node.addInlineFragmentOn('Checkout', (product) => {
+        addFields(product, fields);
+      });
+    });
+  };
+}
```
minasmart: Why is this needed separately from `checkoutQuery`? It looks like there's a lot of shared behaviour.

src/client.js
```diff
@@ -158,7 +160,17 @@ export default class Client {
     });
   }
 
-  fetchCollection(id, query = collectionQuery()) {
+  fetchCollection(id, query = collectionNodeQuery()) {
```
minasmart: Should the queries really vary if they're being requested off node, or off of another object? Should we maybe make an abstraction around fragments, and then share common code between fragments and queries for types?

src/checkout-node-query.js
minasmart: We should probably be striving to request what's required to render a cart and/or a checkout. Is there anything in here that's over and above that?

swalkinshaw: Hard to say 🤔  not sure I feel strongly enough about any of them, so let's keep them

jzjiang: We probably don't need `ready` if we're not using any of the `checkoutComplete` mutations. I'm not sure if we need `paymentDue` if we're not exposing discount/gift card mutations, `order` or `orderStatusUrl`. Otherwise I think the rest could be used.

minasmart: we do need ready. That tells us if taxes have settled out, or if they're still processing. Plus shops need this. We do want to be able to render orders, so I would keep those links.

minasmart: maybe just `orderStatusUrl` for now, since that comes with its own UI.

src/client.js
jzjiang: I'll try to see if I can combine the queries and inline fragments.

minasmart:
looks like this needs a rebase.

test/query-test.js
```diff
@@ -7,7 +7,7 @@ import variantConnectionQuery from '../src/variant-connection-query';
 import optionQuery from '../src/option-query';
 import imageQuery from '../src/image-query';
 import imageConnectionQuery from '../src/image-connection-query';
-import collectionQuery from '../src/collection-query';
+import collectionNodeQuery from '../src/collection-node-query';
```
minasmart: What does `fetchCollections` use now?

jzjiang: It uses `collectionConnectionQuery` to fetch multiple collections off shop and `collectionNodeQuery` to fetch a collection off the root.

src/collection-node-query.js
```diff
@@ -0,0 +1,8 @@
+import imageQuery from './image-query';
+import nodeQuery from './node-query';
+
+const defaultFields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]];
```
minasmart: Are these fields duplicated in `collectionConnectionQuery`? If so, you should  probably share them.

jzjiang: I'll export these fields and update

jzjiang:
Exported default fields in some of the other queries that end up sharing fields (i.e. `image` and `variants`) as well.

minasmart:
👍

jzjiang:
@swalkinshaw do you have any other comments?

PR #341, opened 2017-04-11
[Fix] Use description instead of descriptionPlainSummary

jzjiang:


swalkinshaw:
Shouldn't the base branch be `v1.0alpha`?

minasmart:
I think this needs a rebase.

jzjiang:
Broke some stuff changing the base branch but it should be good now :+1:

PR #343, opened 2017-04-11
[FEATURE] Add methods to fetch collection(s) with products

jzjiang:
Add two methods:
1. `fetchAllCollectionsWithProducts`
2. `fetchCollectionWithProducts`

Images and variants will be paginated on the products returned.

These methods don't have the option to specify fields to query to avoid confusion (e.g. if you start specifying `products` on `fetchCollection` or create a query without `products` and send it to `fetchAllCollectionsWithProducts`)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -57,6 +57,21 @@ function checkoutMutation(type, input, query, client) {
   });
 }
 
+function fetchResourcesForProducts(products, client) {
+  return products.reduce((promiseAcc, product) => {
+    // Fetch the rest of the images and variants for this product
+    promiseAcc.push(client.fetchAllPages(product.images, {pageSize: 250}).then((images) => {
+      product.attrs.images = images;
+    }));
+
+    promiseAcc.push(client.fetchAllPages(product.variants, {pageSize: 250}).then((variants) => {
+      product.attrs.variants = variants;
+    }));
+
+    return promiseAcc;
+  }, []);
+}
+
 /**
  * @class Client
  */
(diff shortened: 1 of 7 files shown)


src/client.js
```diff
@@ -157,6 +157,36 @@ export default class Client {
...
+            product.attrs.images = images;
+          }));
+
+          promiseAcc.push(this.graphQLClient.fetchAllPages(product.variants, {pageSize: 250}).then((variants) => {
+            product.attrs.variants = variants;
+          }));
+
+          return promiseAcc;
+        }, []);
+      }, []);
```
minasmart: This probably shouldn't fetch all products. Fetch the first set, and we expose pagination on the client, so let people paginate. I'd limit this to something much smaller like 20? But at least make it a param on the method.

src/client.js
```diff
@@ -167,6 +197,32 @@ export default class Client {
...
+        promiseAcc.push(this.graphQLClient.fetchAllPages(product.images, {pageSize: 250}).then((images) => {
+          product.attrs.images = images;
+        }));
+
+        promiseAcc.push(this.graphQLClient.fetchAllPages(product.variants, {pageSize: 250}).then((variants) => {
+          product.attrs.variants = variants;
+        }));
+
+        return promiseAcc;
+      }, []);
```
minasmart: Same comment as above.

src/client.js
jzjiang: It fetches images/variants on each product in the collections, not all the products.

minasmart: ok, maybe share this behaviour than? Since it's now used the same way in I think 3 places.

jzjiang:
Since @mina is gone for the week, could you review @lshapton?

src/collection-node-query.js
```diff
@@ -1,8 +1,13 @@
...
 
 export const defaultFields = ['id', 'handle', 'updatedAt', 'title', ['image', imageQuery()]];
 
 export default function collectionNodeQuery(fields = defaultFields) {
   return nodeQuery('Collection', fields);
 }
+
+export function collectionNodeWithProductsQuery() {
+  return nodeQuery('Collection', [...defaultFields, ['products', productConnectionQuery()]]);
+}
```
lshapton: Where is this used? I see `collectionNodeQuery ` is used in `client.js`

jzjiang: I forgot to remove this. Thanks for the catch! :)

PR #346, opened 2017-04-13
Update README

jzjiang:


README.md
```diff
@@ -2,37 +2,146 @@
...
+```
+
+### Creating, Adding and Removing Items from a Checkout
+```javascript
+// Create an empty checkout
+client.createCheckout().then((checkout) => {
+  const addLineItemsInput = {
+    checkoutId: checkout.id,
+    lineItems: [
+      {variantId: 'gid://shopify/ProductVariant/12345', quantity: 5}
```
lshapton: I had to `parseInt(quantity, 10)` and add a radix default value or the linter would complain

jzjiang: I think that only happens because forms convert their input values to strings, so rather than `5` it sends in `"5"`. If we're passing in just numbers, it should work without `parseInt`.

lshapton: Yes, this is true 👍

PR #347, opened 2017-04-13
Support IDs in GID format

swalkinshaw:
This allows for users to pass in an ID both integer and GID format.

### DIFF
--- a/src/node-query.js
+++ b/src/node-query.js
@@ -1,9 +1,9 @@
 import addFields from './add-fields';
-import createGid from './create-gid';
+import normalizeId from './normalize-id';
 
 export default function nodeQuery(type, fields) {
   return function(parentSelection, fieldName, id) {
-    parentSelection.add(fieldName, {args: {id: createGid(type, id)}}, (node) => {
+    parentSelection.add(fieldName, {args: {id: normalizeId(type, id)}}, (node) => {
       node.addInlineFragmentOn(type, (product) => {
         addFields(product, fields);
       });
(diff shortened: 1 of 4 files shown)


test/node-query-test.js
```diff
@@ -0,0 +1,57 @@
...
+  const config = new Config({
+    domain: 'sendmecats.myshopify.com',
+    storefrontAccessToken: 'abc123'
+  });
+
+  const client = new Client(config);
+
+  test('it converts the ID to GID format', () => {
+    const productQuery = nodeQuery('Product', ['title']);
+    const query = client.graphQLClient.query((root) => {
```
swalkinshaw: @jzjiang not sure if there's a more direct way to test this. Technically I can just pass a selection into `productQuery` but wasn't sure if this is the easiest way to do it.

src/node-query.js
```diff
@@ -3,7 +3,16 @@ import createGid from './create-gid';
...
   return function(parentSelection, fieldName, id) {
-    parentSelection.add(fieldName, {args: {id: createGid(type, id)}}, (node) => {
+    const gidProtocol = 'gid:';
+    let gid;
+
+    if (String(id).substr(0, gidProtocol.length) === gidProtocol) {
+      gid = id;
+    } else {
+      gid = createGid(type, id);
+    }
```
minasmart: Maybe pull this into a `normailzeId` function to keep it encapsulated and portable?

swalkinshaw: Will do, makes tests easier as well 👍

swalkinshaw:
Updated. Replaced `createGid` with new `normalizeId`. It didn't seem worth it to me to keep two separate ones still.

swalkinshaw:
No longer needed with opaque IDs.

PR #355, opened 2017-04-20
README + API docs Updates

jzjiang:
Link to Storefront API and make the IDs more accurately reflect what they're supposed to look like.


jzjiang:
Noticed I forgot to rebuild the docs as well so just including that too.

PR #356, opened 2017-04-20
[FEATURE] Migration guide

jzjiang:
- Add link to Migration Guide in README
- Updated table of contents in README
- Add Migration Guide

Closes https://github.com/Shopify/storefront-api-team-projects/issues/59

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # [Shopify](https://www.shopify.com) JavaScript Buy SDK
 [![Circle CI](https://circleci.com/gh/Shopify/js-buy-sdk.png?circle-token=3be0ebe6fbb4841442b86678696947bd4b5456d7)](https://circleci.com/gh/Shopify/js-buy-sdk)
 
+**Note**: For help with migrating from v0.7 of JS Buy SDK to v1 check out the
+[Migration Guide](https://github.com/Shopify/js-buy-sdk/blob/v1.0alpha/docs/MIGRATION_GUIDE.md).
+
 The JS Buy SDK is a lightweight library that allows you to build ecommerce into
 any website. It's based on Shopify's [Storefront API](https://help.shopify.com/api/storefront-api/getting-started)
 and provides the ability to retrieve products and collections from your shop, add products to a cart, and checkout.
@@ -12,8 +15,15 @@ Docs are still a work in progress, but you can view the [API docs](https://githu
 - [Installation](#installation)
 - [Builds](#builds)
 - [Examples](#examples)
+  + [Initializing the Client](#initializing-the-client)
+  + [Fetching Products](#fetching-products)
+  + [Fetching Collections](#fetching-collections)
+  + [Creating, Adding, Updating, and Removing Items from a Checkout](#creating-adding-updating-and-removing-items-from-a-checkout)
+  + [Fetching a Checkout](#fetching-a-checkout)
 - [Example Apps](#example-apps)
+- [Documentation](#documentation)
 - [Contributing](#contributing)
+- [Code of Conduct](#code-of-conduct)
 - [License](#license)
 
 ## Installation
--- a/README.md
+++ b/README.md
@@ -18,7 +18,10 @@ Docs are still a work in progress, but you can view the [API docs](https://githu
   + [Initializing the Client](#initializing-the-client)
   + [Fetching Products](#fetching-products)
   + [Fetching Collections](#fetching-collections)
-  + [Creating, Adding, Updating, and Removing Items from a Checkout](#creating-adding-updating-and-removing-items-from-a-checkout)
+  + [Creating a Checkout](#creating-a-checkout)
+  + [Adding Line Items](#adding-line-items)
+  + [Updating Line Items](#updating-line-items)
+  + [Removing Line Items](#removing-line-items)
   + [Fetching a Checkout](#fetching-a-checkout)
 - [Example Apps](#example-apps)
 - [Documentation](#documentation)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,9 @@
 # [Shopify](https://www.shopify.com) JavaScript Buy SDK
 [![Circle CI](https://circleci.com/gh/Shopify/js-buy-sdk.png?circle-token=3be0ebe6fbb4841442b86678696947bd4b5456d7)](https://circleci.com/gh/Shopify/js-buy-sdk)
 
+**Note**: For help with migrating from v0.7 of JS Buy SDK to v1 check out the
+[Migration Guide](https://github.com/Shopify/js-buy-sdk/blob/v1.0alpha/docs/MIGRATION_GUIDE.md).
+
```
lshapton: I don't think this link works? I get `page not found`

jzjiang: It's set up to link to the v1.0alpha branch, so it should work after this gets merged.

README.md
```diff
@@ -12,8 +15,15 @@ Docs are still a work in progress, but you can view the [API docs](https://githu
 - [Installation](#installation)
 - [Builds](#builds)
 - [Examples](#examples)
+  + [Initializing the Client](#initializing-the-client)
+  + [Fetching Products](#fetching-products)
+  + [Fetching Collections](#fetching-collections)
+  + [Creating, Adding, Updating, and Removing Items from a Checkout](#creating-adding-updating-and-removing-items-from-a-checkout)
```
lshapton: Is there a difference between creating and adding items from a checkout?

jzjiang: It should be creating a checkout and adding line items to a checkout. I'll reword the sections to make that clearer. Thanks for the catch! 😄

swalkinshaw:
Good start 👍 

I'm thinking it would be useful to show a kind of before/after for each example you have now? What do you think? So code in 0.x and then code in 1.x which you already have.

jzjiang:
Yeah I think that's a good idea :+1:

jzjiang:
@minasmart Could you take a look at this? 😄

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+    ```
+    Fields and queries can be nested like so:
+    ```js
+    // Product query with title and variants, where variants have price and title
+    const query = productNodeQuery(['title', ['variants', variantConnectionQuery(['price', 'title'])]]);
+    ```
+    If the `query` argument isn't supplied, a default list of arguments will be used.
+    See the [API reference](https://github.com/Shopify/js-buy-sdk/blob/v1.0alpha/docs/API_REFERENCE.md#Client.Queries) for more details.
+
+3.  `fetchCollection()` and `fetchAllCollections()` no longer return the products in the collections by default.
```
minasmart: `fetchCollection` and `fetchAllCollections` never returned the products. They only returned IDs. This is a net new feature.

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+    To fetch products on the collections, use `fetchCollectionWithProducts()` and `fetchAllCollectionsWithProducts()`,
+    or specify products in the query function parameter.
+    ```js
+    const collectionId = 'Z2lkOi8vc2hvcGlmeS9Db2xsZWN0aW9uLzI1NzY5NzczMQ=='
+    
+    // Use the built-in function
+    client.fetchCollectionWithProducts(collectionId).then((collection) => {
+      console.log(collection); // Collection with all default fields and products with all default fields.
```
minasmart: Show where the products live. It seems obvious to us that `collection.products` is an array, but externally sometimes this stuff isn't obvious.

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+    // or specify a custom query
+    const query = collectionNodeQuery(['title', ['products', productConnectionQuery(['title'])]]);
+
+    client.fetchCollection(collectionId, query).then((collection) => {
+      console.log(collection); // Collection with only title and products with only title
+    });
+    ```
+    Note that `fetchCollectionWithProducts()` and `fetchCollectionsWithProducts()` do not take a `query` argument.
+
+4.  `fetchQueryProducts()` and `fetchQueryCollections()` query different fields and take an optional `query` argument.
```
minasmart: this `query` argument is different than our other query arguments isn't it? Isn't it the shopifyql stuff? If so: make a note that this isn't what people are expecting.

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Carts/Checkouts
+
+Carts are replaced with checkouts. Like the fetch functions, all checkout functions take an optional `query` argument that specifies fields to return on the checkout.
+
+`fetchRecentCart()`, `updateModel()`, and `clearLineItems()` have been removed. `checkoutUrl` has been renamed to `webUrl`.
+
+#### Creating a Checkout
+
+To create a checkout, use `createCheckout()`.
```
minasmart: You should make a note somewhere either here, or around the removed functions, or in both places, that the SDK user is responsible for capturing the ID for later usage. If they want to persist the checkout between sessions, they should store the ID in a cookie or localStorage.

Since this is a migration guide, we should show them how to emulate past behaviour.

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+**v1:**
+```js
+const checkoutId = 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0SW1hZ2UvMTgyMTc3ODc1OTI=';
+
+client.fetchCheckout(checkoutId).then((checkout) => {
+  console.log(checkout); // The retrieved checkout
+});
+```
```
minasmart: Show an example of `fetchRecentCart` and then a v1 version where they use `localStorage.setItem` and `localStorage.getItem` to save and fetch and ID.

docs/MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+  console.log(cart); // Cart with two additional line items
+});
+```
+
+**v1:**
+```js
+const checkoutId = 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0SW1hZ2UvMTgyMTc3ODc1OTI=';
+const lineItems = [
+  {variantId: 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yNTYwMjIzNTk3Ng==', quantity: 5},
+  {variantId: 'Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8yNTYwMjIzNjA0MA==', quantity: 2}
```
minasmart: since there's a new feature around line item attributes, maybe show a second example in here about how to pass them in?

docs/MIGRATION_GUIDE.md
jzjiang: It's kind of confusing because it takes a query object that is the shopifyql stuff and another query arg that's the same as the other query arguments. I'll include an example here to try and make that clearer.

jzjiang:
Updated with your feedback @minasmart!

PR #365, opened 2017-05-24
Remove app id from product listings path

minasmart:
This changes the listings path from `/api/apps/<id>/product_listings` to `/api/product_listings`, since the access token should tell us everything we need to know to resolve an app scope.

The app id is still necessary for know to differentiate local storage carts, in case someone has an implementation from two apps and the same shop on one page (not likely, but it's guarded against for now).


minasmart:
Closing in favour of #453

PR #376, opened 2017-06-19
Remove order fields

millingab:
PR https://github.com/Shopify/shopify/pull/118047 in core is removing some order definition for security and privacy of the merchant and customer. This PR will remove the use of those fields.


millingab:
@minasmart @lyninx Can I get a review on this

PR #381, opened 2017-06-28
bump graphql js client version

lyninx:
Need support from the latest version of `graphql-js-client` to continue work on `buy-button-js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "fb-watchman": "1.9.0",
     "fetch-mock": "5.8.1",
     "fs-extra": "1.0.0",
-    "graphql-js-client": "0.4.2",
+    "graphql-js-client": "0.6.2",
     "graphql-js-schema": "0.6.0",
     "graphql-js-schema-fetch": "1.1.2",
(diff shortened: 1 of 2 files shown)


lyninx:
```AssertionError: src/client.js should pass lint.
1:8 - No default export found in module. import/default``` ???

lyninx:
when i test using `npm test` everything passes, but with yarn i get linting failures.
probably due to this:
`warning "eslint-plugin-shopify@14.0.0" has incorrect peer dependency "eslint@3.3.x".`

minasmart:
Have you tried changing [this line](https://github.com/Shopify/js-buy-sdk/blob/44afd640fe0543b3a245e507a15e64d1463a9874/scripts/rollup-tests.js#L24) to `module: true`, and removing `jsnext: true`?

lyninx:
tried it... but no luck! node resolve doesn't seem to continue to look for jsnext:main even after setting `jsnext: false`. i'll go ahead and ship the graphql-js-client patch to fix it until we figure it out.

PR #383, opened 2017-06-30
Replace addVariants with createLineItemsFromVariants because its deprecated

pvomhoff:
I just followed the Getting started and the following error message comes up:
`[JS-BUY-SDK]:  CartModel -  addVariants is deprecated, please use createLineItemsFromVariants instead`

So the Getting started includes an deprecated version. I replaced it by `addVariants`.


swalkinshaw:
Good catch, thanks 👍

PR #385, opened 2017-07-10
[fix] CircleCI badge appropriate for branch

doublerebel:
Hello,

I see this library is passing its tests for this branch, which unfortunately was not reflected in the README.  I noticed when I downloaded and saw the tests passing locally.  This fixes up the badge and link to point to v1.0alpha branch.


minasmart:
Thanks for the fix!

PR #387, opened 2017-07-12
backwards compatibility support for old client configuration

lyninx:
in the context of buy-button-js: the previous config process was to call `ShopifyBuy.buildClient(config)`. the current 1.0 version no longer has that method, and asks you to create an instance of ShopifyBuy with the config passed to the constructor instead. 

I've created a `.buildClient` wrapper method which adapts the old configuration object {apiKey, domain, appId}, to the new `Config` object {domain, storefrontAccessToken}. Further changes need to be made in buy-button-js in order to keep existing buy button embeds alive as well, which are linked in the buy-button-js PR below.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -65,6 +65,14 @@ function fetchResourcesForProducts(products, client) {
   }, []);
 }
 
+function adaptConfig(config) {
+  const accessToken = config.storefrontAccessToken || config.apiKey;
+  return new Config({
+    domain: config.domain,
+    storefrontAccessToken: accessToken,
+  });
+}
+
 /**
  * The JS Buy SDK Client.
  * @class
@@ -115,6 +123,14 @@ class Client {
       variantQuery,
       checkoutNodeQuery
     };
+  }
+
+  /**
+   * A wrapper around the constructor, for backwards compatability.
+   */
+  static buildClient(config) {
+    const newConfig = adaptConfig(config);
+    return new Client(newConfig);
   }
 
   /**
--- a/src/client.js
+++ b/src/client.js
@@ -66,8 +66,11 @@ function fetchResourcesForProducts(products, client) {
   }, []);
 }
 
-function adaptConfig(config) {
-  const accessToken = config.accessToken || config.apiKey;
+function normalizeAccessToken(config) {
+  if (config.accessToken) {
+    console.warn("[ShopifyBuy] accessToken is deprecated as of v1.0, please use storefrontAccessToken instead.")
+  }
+  const accessToken = config.storefrontAccessToken || config.accessToken;
 
   return new Config({
     domain: config.domain,
@@ -131,7 +134,7 @@ class Client {
    * A wrapper around the constructor, for backwards compatability.
    */
   static buildClient(config) {
-    const newConfig = adaptConfig(config);
+    const newConfig = normalizeAccessToken(config);
 
     return new Client(newConfig);
   }
--- a/src/client.js
+++ b/src/client.js
@@ -119,7 +119,8 @@ class Client {
   }
 
   /**
-   * A wrapper around the constructor, for backwards compatability.
+   * Primary entry point for building a new Client.
+   * @namespace
    */
   static buildClient(config) {
     const newConfig = new Config(config);
--- a/src/client.js
+++ b/src/client.js
@@ -1,4 +1,5 @@
 import GraphQLJSClient from './graphql-client';
+import Config from './config';
 import types from '../types';
 import productNodeQuery from './product-node-query';
 import productConnectionQuery from './product-connection-query';
@@ -66,10 +67,11 @@ function fetchResourcesForProducts(products, client) {
 }
 
 function adaptConfig(config) {
-  const accessToken = config.storefrontAccessToken || config.apiKey;
+  const accessToken = config.accessToken || config.apiKey;
+
   return new Config({
     domain: config.domain,
-    storefrontAccessToken: accessToken,
+    storefrontAccessToken: accessToken
   });
 }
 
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -65,6 +65,14 @@ function fetchResourcesForProducts(products, client) {
   }, []);
 }
 
+function adaptConfig(config) {
+  const accessToken = config.storefrontAccessToken || config.apiKey;
```
swalkinshaw: It's just `accessToken`

src/client.js
```diff
@@ -65,6 +66,15 @@ function fetchResourcesForProducts(products, client) {
   }, []);
 }
 
+function adaptConfig(config) {
```
minasmart: Since you're only normalizing one field, why not just have a `normalizeAccessToken` method? that receives the config, and returns whichever token, as well as logging a deprecation notice for the deprecated config value.

That way, the transformation isn't as opaque in the build client method.

Also, wasn't there some resolution about this being only `accessToken`? @swalkinshaw can you fill me in with what happened here?

swalkinshaw: What's there now is correct. It was `storefrontAccessToken` before which was incorrect.

minasmart: I'm confused. It's returning `storefrontAccessToken: accessToken`

swalkinshaw: Before it was:

```js
const accessToken = config.storefrontAccessToken || config.apiKey;
```

minasmart: ah, k.

src/client.js
```diff
@@ -65,6 +66,18 @@ function fetchResourcesForProducts(products, client) {
...
+function normalizeAccessToken(config) {
+  if (config.accessToken) {
+    console.warn("[ShopifyBuy] accessToken is deprecated as of v1.0, please use storefrontAccessToken instead.")
+  }
+  const accessToken = config.storefrontAccessToken || config.accessToken;
+
+  return new Config({
+    domain: config.domain,
+    storefrontAccessToken: accessToken
+  });
```
minasmart: The config should be built in build client. This should really just return the value that needs normalization.

minasmart: Doing it this way unnecessarily obfuscates construction of the config.

minasmart: Also, this normalization should probably all be happening in the config constructor. (sorry for all the comments)

lyninx: no worries. agreed, it should just happen in the constructor.

src/config.js
```diff
@@ -1,3 +1,12 @@
+
+function normalizeAccessToken(config) {
+  // eslint-disable-next-line
+  console.warn('[ShopifyBuy] accessToken is deprecated as of v1.0, please use storefrontAccessToken instead.');
+  const accessToken = config.storefrontAccessToken || config.accessToken;
```
swalkinshaw: Is there a point in this anymore?

This function is only called when the config has an `accessToken` key.

Might as well just handle this directly below:

```js
 constructor(attrs) {
    this.requiredProperties.forEach((key) => {
      if (attrs.hasOwnProperty(key)) {
        this[key] = attrs[key];
      } else if (key === 'accessToken') {
        console.warn('[ShopifyBuy] accessToken is deprecated as of v1.0, please use storefrontAccessToken instead.');
        this.storefrontAccessToken = attrs['accessToken'];
      } else {
        throw new Error(`new Config() requires the option '${key}'`);
      }
    });
  }
```

lyninx: ah, good point. the function is pretty tightly coupled with that block anyway, so i'll just make it inline.

src/config.js
```diff
@@ -28,6 +28,10 @@ class Config {
     this.requiredProperties.forEach((key) => {
       if (attrs.hasOwnProperty(key)) {
         this[key] = attrs[key];
+      } else if (key === 'accessToken') {
+        // eslint-disable-next-line
+        console.warn('[ShopifyBuy] accessToken is deprecated as of v1.0, please use storefrontAccessToken instead.');
```
minasmart: disable the specific rule.

src/client.js
```diff
@@ -118,6 +119,15 @@ class Client {
   }
 
   /**
+   * A wrapper around the constructor, for backwards compatability.
```
minasmart: this isn't for backward compatibility. This is the primary entry point for most consumers.

minasmart: also, this should be in jsdoc format.

lyninx: it is for most of the current consumers, but i assumed that since it's no longer the recommended configuration method, we're still supporting the old method, making it backwards compatible. words are hard.

minasmart: Even the artifact we build exports the single `ShopifyBuy` global, and `buildClient` is the entry point.

lyninx: so, what's the point of having both methods? why have `.buildClient` versus passing in a `Config` object to `ShopifyBuy`?

minasmart: Do you mean passing a `Config` object to `Client` class?

Most vanilla javascript can't handle the `import` keyword, so we need a globally accessible entry point. For the purposes of compilation, it's nice to have a single entry point as a file so we can take advantage of tree shaking. And we have one external interface, so that if we need to change things internally, we can just change the implementation of `buildClient` to fit what lives under the hood.

Does that answer what you're asking? I'm not sure I understand.

lyninx: yep. that clears things up. why not just use `.buildClient` in our examples then?

minasmart: We should.

lyninx: OK 👍

src/client.js
```diff
@@ -118,6 +119,16 @@ class Client {
   }
 
   /**
+   * Primary entry point for building a new Client.
+   * @namespace
```
minasmart: is `@namespace` the right directive here?

PR #403, opened 2017-08-23
fix deprecated Config attrs + add tests

lyninx:
as it turns out, the last time I updated this deprecated properties on the Config class, it didn't actually do anything. i've rewritten it, and added tests, so we should be good now.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -19,19 +19,36 @@ class Config {
   }
 
   /**
+   * Deprecated properties that map directly to required properties
+   * @attribute deprecatedProperties
+   * @default ['accessToken', 'apiKey']
+   * @type Object
+   * @private
+   */
+  get deprecatedProperties() {
+    return {
+      accessToken: 'storefrontAccessToken',
+      apiKey: 'storefrontAccessToken'
+    };
+  }
+
+  /**
    * @constructs Config
    * @param {Object} attrs An object specifying the configuration. Requires the following properties:
    *   @param {String} attrs.storefrontAccessToken The {@link https://help.shopify.com/api/reference/storefront_access_token|Storefront access token} for the shop.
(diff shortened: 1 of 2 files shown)


test/config-test.js
```diff
@@ -24,6 +24,30 @@ suite('config-test', () => {
     });
   });
 
+  test('it assigns apiKey as storefrontAccessToken', () => {
```
swalkinshaw: small thing about testing, I usually prefer to avoid variables in a case like this and do it more explicitly/directly. It makes it a little easier to quickly see what's going on. Here's the alternative:

```js
test('it assigns apiKey as storefrontAccessToken', () => {
  const config = new Config({
    domain: 'website.com',
    apiKey: 'i am actually a storefrontAccessToken'
  });

  assert.equal(config.storefrontAccessToken, 'i am actually a storefrontAccessToken');
});
```

src/config.js
```diff
@@ -19,19 +19,36 @@ class Config {
...
    * @constructs Config
    * @param {Object} attrs An object specifying the configuration. Requires the following properties:
    *   @param {String} attrs.storefrontAccessToken The {@link https://help.shopify.com/api/reference/storefront_access_token|Storefront access token} for the shop.
    *   @param {String} attrs.domain The `myshopify` domain for the shop (e.g. `graphql.myshopify.com`).
    */
   constructor(attrs) {
+    Object.keys(this.deprecatedProperties).forEach((key) => {
+      if (!attrs.hasOwnProperty(key)) { return; }
+      // eslint-disable-next-line no-console
+      console.warn(`[ShopifyBuy] ${key} is deprecated as of v1.0, please use ${this.deprecatedProperties[key]} instead.`);
```
minasmart: Maybe say `Config property ${key} is ...`, just so it's more specific, instead of just a bare word

PR #405, opened 2017-08-30
Converts all query builders to static queries

minasmart:
Part of the larger beta / release overhaul. Follow along in #406


src/handle-checkout-mutation.js
```diff
@@ -0,0 +1,17 @@
+export default function handleCheckoutMutation(mutationRootKey, client) {
```
swalkinshaw: Is this file used anywhere?

minasmart: yep. in `client.js`

PR #410, opened 2017-09-08
Move integration tests into their own files and test the right things

minasmart:
This moves integration tests out of the `client-test` suite, and into their own test suites. It also makes it so that we're testing the thing we can test: resolution patterns, instead of validating the data in fixtures.

This PR also normalizes behaviours in resolvers in the public client methods.


chrisbutcher:
It might be wise for me to join on that pairing too :)

test/fetch-resources-for-products-test.js
```diff
@@ -0,0 +1,64 @@
+import assert from 'assert';
+import fetchResourcesForProducts from '../src/fetch-resources-for-products';
+
+suite('fetch-resources-for-product-test', () => {
+  test('it fetches all pages of images and for a single product', () => {
+    const imagesArray = ['images'];
+    const variantsArray = ['variants'];
+    const fixture = {
+      images: imagesArray,
```
chrisbutcher: Where are we using `fixture.images` and `.variants`? I'm only seeing them access via `.attrs`.

test/default-resolver.js
```diff
@@ -0,0 +1,3 @@
+export default function defaultResolver(path) {
```
chrisbutcher: Are we using this definition of defaultResolver anywhere? I see us importing only `src/default-resolver` elsewhere.

test/fetch-resources-for-products-test.js
minasmart: because the test shouldn't guarantee implementation. If this goes through anything that has expectations of model shape, it will expect the top level proxy fields. `GraphModel` has an internall `attrs` hash, and top level fields may be proxies, or overrides.

test/default-resolver.js
minasmart: yes. It's used extensively in `client.js`

PR #420, opened 2017-09-20
Clean up build and use modern rollup watch

minasmart:
Removes custom build and watch scripts. Uses basic rollup configs and rollup watch. Uses babel preset env to target browsers and node.


rollup-browser-tests.config.js
```diff
@@ -0,0 +1,57 @@
...
+  builtins(),
+  nodeResolve({
+    jsnext: true,
+    main: true,
+    preferBuiltins: false
+  }),
+  babel({
+    babelrc: false,
+    presets: [
+      [`${process.cwd()}/node_modules/babel-preset-env/lib/index`, {
```
swalkinshaw: Any reason not to keep these in `.babelrc`? Wouldn't it avoid the weirdness of having to specify the full path?

minasmart: It's different between builds and using babel env has quirks.

Issue #425, opened 2017-09-29
[1.0beta] productResource.fetchAll generates an invalid query


agos:
Using `productResource.fetchAll()` results in a query that errors, indipendent of the passed `pageSize` parameter.

Here is the generated query:

```GraphQL
fragment VariantFragment on ProductVariant {
  id
  title
  price
  weight
  available
  image {
    id
    src
    altText
  }
  selectedOptions {
    name
    value
  }
}

fragment ProductFragment on Product {
  id
  createdAt
  updatedAt
  descriptionHtml
  description
  handle
  productType
  title
  vendor
  tags
  publishedAt
  onlineStoreUrl
  options {
    id
    name
    values
  }
  images(first: 250) {
    pageInfo {
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        src
        altText
      }
    }
  }
  variants(first: 250) {
    pageInfo {
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        ...VariantFragment
      }
    }
  }
}

query ($first: Int!, $query: String, $sortKey: ProductSortKeys, $reverse: Boolean) {
  shop {
    products(first: $first, query: $query, sortKey: $sortKey, reverse: $reverse) {
      pageInfo {
        hasNextPage
        hasPreviousPage
      }
      edges {
        cursor
        node {
          ...ProductFragment
        }
      }
    }
  }
}
```

here is the accompaining `variables` field:
```json
{
  "pageSize": 20
}
```

and here is the error response:

```json
{
  "errors": [
    {
      "message": "Variable first of type Int! was provided invalid value",
      "locations": [{ "line": 1, "column": 545 }],
      "value": null,
      "problems": [
        { "path": [], "explanation": "Expected value to not be null" }
      ]
    }
  ]
}
```

It appears that parameters `first`, `query`, `sortKey`, `reverse` are expected (see [source](https://github.com/Shopify/js-buy-sdk/blob/v1.0.0-beta.1/src/graphql/productConnectionQuery.graphql)), but only `pageSize` is provided.

minasmart:
Fixed in #427

Issue #426, opened 2017-09-29
[1.0beta] productResource.fetchQuery generates an invalid query


agos:
Using `productResource.fetchQuery()` results in a query that errors, indipendent of the passed parameters.

Here is the generated query:

```GraphQL
fragment VariantFragment on ProductVariant {
  id
  title
  price
  weight
  available
  image {
    id
    src
    altText
  }
  selectedOptions {
    name
    value
  }
}

fragment ProductFragment on Product {
  id
  createdAt
  updatedAt
  descriptionHtml
  description
  handle
  productType
  title
  vendor
  tags
  publishedAt
  onlineStoreUrl
  options {
    id
    name
    values
  }
  images(first: 250) {
    pageInfo {
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        src
        altText
      }
    }
  }
  variants(first: 250) {
    pageInfo {
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        ...VariantFragment
      }
    }
  }
}

query ($first: Int!, $query: String, $sortKey: ProductSortKeys, $reverse: Boolean) {
  shop {
    products(first: $first, query: $query, sortKey: $sortKey, reverse: $reverse) {
      pageInfo {
        hasNextPage
        hasPreviousPage
      }
      edges {
        cursor
        node {
          ...ProductFragment
        }
      }
    }
  }
}
```

here is the accompaining `variables` field:
```json
{
  "first": 20,
  "sortKey": { "key": "ID" }
}
```

and here is the error response:

```json
{
  "errors": [
    {
      "message": "Variable sortKey of type ProductSortKeys was provided invalid value",
      "locations": [
        {
          "line": 1,
          "column": 571
        }
      ],
      "value": {
        "key": "ID"
      },
      "problems": [
        {
          "path": [],
          "explanation": "Expected \"{\\\"key\\\"=\u003e\\\"ID\\\"}\" to be one of: TITLE, PRODUCT_TYPE, VENDOR, UPDATED_AT, CREATED_AT, ID, RELEVANCE"
        }
      ]
    }
  ]
}
```

minasmart:
Fixed in #427

gsalgadotoledo:
This issue is still happening in 2018, I corrected setting the first value using this way `new Number(250)`

minasmart:
Can you show an example of the query that's failing?

gsalgadotoledo:
```
[...]

PR #427, opened 2017-09-29
Fix formats of data sent to queries

minasmart:
Also a tiny fix around using `pageSize` and `first` interchangeably. Be consistent with relay and use `first` always.


minasmart:
Fixes #425 & #426

PR #441, opened 2017-10-31
Upgrade graphql compiler and client to fix underlying pagination issues

minasmart:
This upgrades the graphql-js-client-compiler to hoist query variables, instead of dynamically defining them at every usage.

It also upgrades the graphql-js-client version to handle generated pagination queries with variables and named fragments (which were previously being ignored in generated pagination queries).

This is hopefully the last thing that needs to be addressed before `v1.0` gets launched.


src/graphql/CollectionsProductsFragment.graphql
```diff
@@ -1,5 +1,5 @@
 fragment CollectionsProductsFragment on Collection {
-  products(first: $productsFirst) {
+  products(first: 20) {
```
swalkinshaw: we want to hardcode this value?

minasmart: fragments have no access to variables.

minasmart: This wasn't initially valid, it just passed the parser somehow.

I'm not sure how many layers deep we want to parameterize things. Ideally, if people want heavy customization, people should be using the storefront API directly? But this is totally just opinion.

harisaurus: is there a default that shopify uses across other places? if not, 20 seems reasonable to me

minasmart: There is on some of our APIs. The point of the change here was to get stuff to compile.

I can open these params up later by taking this out of a fragment.

PR #450, opened 2017-11-17
V1.0beta Supporting customer auth/management

ericpoon:
This enhancement supports the following:

1. Create a new customer(user) account
2. Update the customer(user) profile
3. Create a customer access token for an existing customer (User authentication)
4. Update/delete the customer access token
5. Create/update/remove an address for an existing customer
6. Set the default address for an existing customer
7. Associate/disassociate a checkout with an existing customer
8. Send a reset password email to the customer, as the first step in the reset password process.

Please refer to docs/API_REFERENCE.md for details.


minasmart:
This functionality is too specific to the use case to exist in this SDK. It also increases the payload size of the SDK significantly. Lastly, there are no tests. If you wanted to maintain this as an add on, or in your own personal fork though, 👍 

Thanks for your work!

PR #452, opened 2017-11-24
allow someone to pass a fetch function (like in node)

minasmart:
Expose an option to provide a fetch-compatible function to the `buildClient` method, so it's easier to integrate into node projects

Docs to come shortly

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -25,29 +25,45 @@ class Client {
   /**
    * Primary entry point for building a new Client.
    */
-  static buildClient(config) {
+  static buildClient(config, fetchFunction) {
     const newConfig = new Config(config);
 
-    return new Client(newConfig);
+    return new Client(newConfig, GraphQLJSClient, fetchFunction);
   }
 
   /**
    * @constructs Client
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -25,29 +25,45 @@ class Client {
...
+      'X-SDK-Version': version,
+      'X-Shopify-Storefront-Access-Token': config.storefrontAccessToken
+    };
+
+    if (fetchFunction) {
+      headers['Content-Type'] = 'application/json';
+      headers.Accept = 'application/json';
+
+      this.graphQLClient = new GraphQLClientClass(types, {
+        fetcher: function fetcher(graphQLParams) {
```
swalkinshaw: What about modifying `httpFetcher` to accept a `fetchFunction` as well?

graphql-js-client has the same issue right now I assume? It would allow both to be fixed.

minasmart: I thought about that, but I think I like graphql js client to be more open. It accepts an `httpFetcher` function that just receives GraphQL params. I think in the wider node world, we shouldn't assume `fetch` as an API, but in this one library it's probably a safer, but still less-ideal hedge.

PR #455, opened 2017-12-05
Add minified UMD version for CDN distribution

minasmart:
Also includes a node in the readme


minasmart:
@swalkinshaw it doesn't. If you pull out the `.min` part of the URI, you'll see the unminified version

PR #484, opened 2018-02-26
Switch default resolver to return promise and reject undefined keys with proper error messages

StefanSlehta:
There were issues where the resolver was failing when trying to fetch products with invalid `ids` . This PR fixes it and rejects with proper error messages

### DIFF
--- a/src/default-resolver.js
+++ b/src/default-resolver.js
@@ -1,9 +1,23 @@
+export const defaultErrors = [{message: 'an unknown error has occured'}];
+
 export default function defaultResolver(path) {
   const keys = path.split('.');
 
-  return function({model}) {
-    return keys.reduce((ref, key) => {
-      return ref[key];
-    }, model);
+  return function({model, errors}) {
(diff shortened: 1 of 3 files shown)


src/default-resolver.js
```diff
@@ -1,9 +1,23 @@
+export const defaultErrors = [{message: 'an unknown error has occured'}];
```
minasmart: To be consistent with our APIs this error should end in a period.

test/default-resolver-test.js
```diff
@@ -23,6 +25,37 @@ suite('default-resolver-test', () => {
...
+    return resolve({errors}).then(() => {
+      assert.ok(false, 'should not resolve');
+    }).catch((resolvedErrors) => {
+      assert.equal(resolvedErrors, errors);
+
+      return true;
+    });
+  });
+
+  test('it rejects with synthetic errors hash if no top level key available', () => {
```
minasmart: `it rejects with a synthetic errors hash`

PR #485, opened 2018-03-07
Return a product image usable in the thumbnail

StefanSlehta:
Addressing [#2336](https://github.com/Shopify/buy-button/issues/2336)


minasmart:
When you have a chance, this resolves a number of outstanding issues on this repository, if you could close and reference this PR, that would help a lot.

PR #506, opened 2018-04-09
Reject checkout mutation if graphql errors are returned

StefanSlehta:


src/handle-checkout-mutation.js
```diff
@@ -1,8 +1,14 @@
 export default function handleCheckoutMutation(mutationRootKey, client) {
-  return function({data, model}) {
+  return function({data, errors, model}) {
     const rootData = data[mutationRootKey];
     const rootModel = model[mutationRootKey];
-    const userErrors = rootData.userErrors;
+    let userErrors = [];
+
+    if (errors != null && errors.length) {
```
minasmart: don't use coercive equals, but also `errors && errors.length` is sufficient. We don't care if it's `null` or `undefined` or `0`. we just care taht it's falsey

src/handle-checkout-mutation.js
```diff
@@ -1,8 +1,14 @@
...
+      userErrors = errors;
```
minasmart: just call reject in each branch. no need to assign names.

`if (errors && errors.length) {
  return Promise.reject(...) {
} else if (rootData && rootData.userErrors && rootData.userErrors.length)
  return Promise.reject(...) {
}

src/handle-checkout-mutation.js
swalkinshaw: This doesn't necessarily mean a mutation "failed". The mutation can succeed but an error could be caused by a return field. Depends what you want, but you could check if `rootData` is `null` as well.

src/handle-checkout-mutation.js
```diff
@@ -1,15 +1,19 @@
...
-    const userErrors = rootData.userErrors;
 
-    if (userErrors.length) {
-      return Promise.reject(new Error(JSON.stringify(userErrors)));
+    if (!rootData) {
+      if (errors && errors.length) {
+        return Promise.reject(new Error(JSON.stringify(errors)));
+      }
+    } else if (rootData.userErrors && rootData.userErrors.length) {
+      return Promise.reject(new Error(JSON.stringify(rootData.userErrors)));
```
minasmart: This may be more robust as

```javascript
if (rootData && rootData.checkout) {
  return client.fetchAllPages(rootModel.checkout.lineItems, {pageSize: 250}).then((lineItems) => {
    rootModel.checkout.attrs.lineItems = lineItems;
    rootModel.checkout.errors = errors;
    rootModel.checkout.userErrors = rootModel.userErrors;

    return rootModel.checkout;
  });
}

if (errors && errors.length) {
  return Promise.reject(new Error(JSON.stringify(errors)));
}

if (rootData && rootData.userErrors && rootData.userErrors.length) {
  return Promise.reject(new Error(JSON.stringify(rootData.userErrors)));
}

return Promise.reject(new Error(`The ${mutationRootKey} mutation failed due to an unknown error.`));
```

Issue #498, opened 2018-03-27
Update checkout object's customAttributes using mutations - Express JS


abhilash29:
I'm currently using Express JS. I've successfully integrated the JS buy SDK and I'm successfully able to create/update/delete checkouts and their lineItems.
I also managed to create customAttributes during checkout creation.
However, what I'm not able to figure out is the correct syntax/code to use mutations as mentioned in [checkoutAttributesUpdate][1] link .
Unfortunately, there doesn't seem to be an out of the box function to do this .

This is how I added customAttributes.

   
```
const input = {
    customAttributes: customAttributes
};
client.checkout.create(input).then((checkout) => {
    //Remaining stuff
  
});

```
  [1]: https://help.shopify.com/api/storefront-api/reference/mutation/checkoutattributesupdate#interactive-example

AmeliaBR:
I'm having the same issue, and this seems like a significant gap in the API to me.

As far as I can tell, any time I need to update cart attributes, I either need to 

- create a new checkout with a copy of all existing data (e.g. line items) about the cart.
- process the GraphQL mutation myself

An `update` method in the [`CheckoutResource` class](https://github.com/Shopify/js-buy-sdk/blob/master/src/checkout-resource.js]) would be very useful.

thohot303:
Unbelievable that Shopify just updated the Buy Button implementation without having proper replacement methods in place. I can now no longer update custom attributes via permalinks and an update method on the CheckoutResource object is not available.

StefanSlehta:
You can now use the [update](https://github.com/Shopify/js-buy-sdk/blob/master/src/checkout-resource.js#L89) method  in version `1.5.0` to add manipulate `customAttributes` on a `Checkout`

thohot303:
Thanks, much appreciated. When will this be moved to the current implementation, though?

StefanSlehta:
The only thing that's been changed in `1.6.0` was the addition of this method, so you can update to the newest version without breaking any functionality.

thohot303:
My app requires users to install & modify the Shopify-generated Buy Button code.
The [script URL](https://sdks.shopifycdn.com/buy-button/latest/buy-button-storefront.min.js) points to version **1.4.0**.

StefanSlehta:
It's published now, so the `1.6.0` should be loading the

thohot303:
Awesome, works like a charm now, thanks!

Issue #384, opened 2017-07-08
Updating checkout attributes after checkout creation


doublerebel:
Hello,

There is a clear way to [create a checkout with many attributes](https://github.com/Shopify/js-buy-sdk/blob/v1.0alpha/docs/API_REFERENCE.md#clientcreatecheckoutinput-query--promise--graphmodel), but it appears the only APIs exposed for updating a checkout after creation are add/removeLineItem.

I have a common use case, a user creates a checkout while not logged in (or registered), and later logs in before purchase.  Therefore I need to update the email and add custom attribute of user id.  Is there a way to do this easily with the lower-level GraphQL client?

swalkinshaw:
I'd check out the available mutations in our [docs](https://help.shopify.com/api/storefront-api/reference/mutation). Any of these can be use via the GraphQL client although it may make sense to expose some more to this library.

You may want:

* https://help.shopify.com/api/storefront-api/reference/mutation/checkoutemailupdate
* https://help.shopify.com/api/storefront-api/reference/mutation/checkoutcustomerassociate
* https://help.shopify.com/api/storefront-api/reference/mutation/checkoutattributesupdate

doublerebel:
Thanks for pointing me in the right direction, I'll check these out.

youanden:
@minasmart I'd like to use the `checkoutAttributesUpdate` mutation to update the cart note.

It's not available in the SDK yet as far as I can see. I also can't get it working with the unsupported permalinks like
```https://checkout.shopify.com/{storeid}/checkouts/{cartid}?key={key}&note=blabla``` [found here](https://help.shopify.com/themes/customization/cart/use-permalinks-to-preload-cart#what-is-a-cart-permalink)

Is there a place to vote for next sdk implementations? Right now @pulber has manually ported the sdk and augmented it with an `updateCheckoutAttributes` function for our internal use.

If it was easier to just pass a string to the graphql api at a lower level like @doublerebel requested  we wouldn't need it. It also seems like there is a lot of code to write in order to implement a mutation available to the api. 

I guess something like `ShopifyAPI.clientInstance.graphQLClient.send` but lower level, passing a string instead of an object and then calling `toString()`.

minasmart:
Hey there! That mutation was added after the initial release of this SDK. It will be added at some point in the near future.

[...]

PR #541, opened 2018-06-18
Add support for checkout update attributes

StefanSlehta:
Adds support for updating checkout attributes. Closing #498

### DIFF
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -69,11 +69,11 @@ class CheckoutResource extends Resource {
   }
 
   /**
-   * Updates a checkout.
+   * Replaces the value of checkout's custom attributes and/or note with values defined in the input
    *
    * @example
    * const checkoutId = 'Z2lkOi8vc2hvcGlmeS9DaGVja291dC9kMTZmM2EzMDM4Yjc4N=';
-   * const input = { customAttributes: [{ key: "MyKey", value: "MyValue" }] };
+   * const input = {customAttributes: [{key: "MyKey", value: "MyValue"}]};
    *
    * client.checkout.update(checkoutId, input).then((checkout) => {
    *   // Do something with the updated checkout
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -8,6 +8,7 @@ import checkoutCreateMutation from './graphql/checkoutCreateMutation.graphql';
 import checkoutLineItemsAddMutation from './graphql/checkoutLineItemsAddMutation.graphql';
 import checkoutLineItemsRemoveMutation from './graphql/checkoutLineItemsRemoveMutation.graphql';
 import checkoutLineItemsUpdateMutation from './graphql/checkoutLineItemsUpdateMutation.graphql';
+import checkoutAttributesUpdateMutation from './graphql/checkoutAttributesUpdateMutation.graphql';
 
 /**
  * The JS Buy SDK checkout resource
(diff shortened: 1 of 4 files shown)


src/checkout-resource.js
```diff
@@ -67,6 +68,30 @@ class CheckoutResource extends Resource {
       .then(handleCheckoutMutation('checkoutCreate', this.graphQLClient));
   }
 
+  /**
+   * Updates a checkout.
```
minasmart: does this update and persist new values sent, or merge with existing values?

StefanSlehta: It updates with the new values sent. So if previously `customAttributes: [{key: "Key', value: "Value"}]` existed, and the new input was `customAttributes: [{key: "DifferentKey", value: "DifferentValue"}]`, the new value would be persisted, and the old discarded.

rebeccajfriedman: Shouldn't this be `Update's a checkout's custom attributes` or something more specific?

minasmart: If call one was `customAttributes: [{key: "keyOne", value: "valueOne"}]` and call two was `customAttributes: [{key: "keyTwo", value: "valueTwo"}]`

Would the resultant `customAttributes` have both `keyOne` and `keyTwo` or just `keyTwo`? If it's the latter, `replaces checkout custom attributes` would be the better descriptor. Either way, `updates a checkout` is way to vague when we're specifically talking about a checkout's custom attributes.

StefanSlehta: You can also update `allowPartialAddress` and `note`. Maybe it can be [checkoutAttributesUpdate](https://help.shopify.com/api/custom-storefronts/storefront-api/reference/input_object/checkoutattributesupdateinput )

rebeccajfriedman: The description on the graphql side is `Updates the attributes of a checkout.`

minasmart: which is also vague. Line items are attributes of a checkout, but they're not updatable here. Also, what was the answer to my question 3 comments up @StefanSlehta ?

StefanSlehta: The resultant value of `customAttributes` would be just `keyTwo`, however, you can still update only a `note`, without providing `customAttributes` and it wouldn't delete the ones that are currently on the `checkout`. Maybe we could say `Updates or replaces checkout's attributes ` ?

minasmart: we should probably be more specific than just `a checkout's attributes`. this specifically updates `a checkouts `customAttributes and/or notes` field. right? attributes that exist on checkout are significantly more than what can be updated wit this mutation.

minasmart: updates or replaces is ambiguous. this explicitly replaces either one or both of those fields.

StefanSlehta:
Does the new description work?

src/checkout-resource.js
```diff
@@ -67,6 +68,30 @@ class CheckoutResource extends Resource {
       .then(handleCheckoutMutation('checkoutCreate', this.graphQLClient));
   }
 
+  /**
+   * Replaces the value of checkout's custom attributes and/or note with values defined in the input
+   *
+   * @example
+   * const checkoutId = 'Z2lkOi8vc2hvcGlmeS9DaGVja291dC9kMTZmM2EzMDM4Yjc4N=';
+   * const input = { customAttributes: [{ key: "MyKey", value: "MyValue" }] };
```
minasmart: the example should fit the rest of the examples, no spaces after or before `{` and `}`

warrenmcquinn:
Wow! Does this mean that #384 has an answer?

StefanSlehta:
@warrenmcquinn Yes, I [commented](https://github.com/Shopify/js-buy-sdk/issues/384#issuecomment-398401103) on that issue as well.

warrenmcquinn:
@StefanSiehta -- thank you for your awesome contribution! And thanks to the whole Shopify dev team. You just enabled our exact use-case, so we're moving forward with our Shopify integration.

PR #542, opened 2018-06-18
Change method name to something more specific

StefanSlehta:


warrenmcquinn:
Hey there, looks like the method name in the example wasn't updated (`update` --> `updateAttributes`)! I'd make a PR but it's such a minor change. Thanks!

PR #557, opened 2018-08-21
checkoutEmailIUpdate mutation

jayjetley:
This PR adds support to the checkoutEmailUpdate mutation. I have also added email to the checkout object.

https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/mutation/checkoutemailupdate

I have not updated the documents as I noticed in other PRs you've requested not to update as it causes merge conflicts but can do if you would like.

### DIFF
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -106,7 +106,7 @@ class CheckoutResource extends Resource {
    * });
    *
    * @param {String} checkoutId The ID of the checkout to add discount to.
-   * @param {String} email The discount code to apply to the checkout.
+   * @param {String} email The email address to apply to the checkout.
    * @return {Promise|GraphModel} A promise resolving with the updated checkout.
    */
   updateEmail(checkoutId, email) {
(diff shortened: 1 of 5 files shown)


src/checkout-resource.js
```diff
@@ -93,6 +94,27 @@ class CheckoutResource extends Resource {
...
+   * @example
+   * const checkoutId = 'Z2lkOi8vc2hvcGlmeS9DaGVja291dC9kMTZmM2EzMDM4Yjc4N=';
+   * const email = 'user@example.com';
+   *
+   * client.checkout.updateEmail(checkoutId, email).then((checkout) => {
+   *   // Do something with the updated checkout
+   * });
+   *
+   * @param {String} checkoutId The ID of the checkout to add discount to.
+   * @param {String} email The discount code to apply to the checkout.
```
StefanSlehta: Can you update this please?

jayjetley:
I have updated the comment to reflect the parameter. Damn that copy and paste :-)

StefanSlehta:
@jayjetley We just [published](https://www.npmjs.com/package/shopify-buy)  these changes  so you can start using them right away if you need to. Thanks again!

jayjetley:
Hi Stefan,

Many thanks for this. Does this get automatically published to npm?

Thanks

Jay

ahrenpradhan:
Thanks @jayjetley 
But I would reccommend if we could add this feature in docs as well

PR #565, opened 2018-10-12
Handle a null response when fetching a checkout

### LINKED ISSUE #563: TypeError: Cannot read property 'lineItems' of null
**Describe the bug**
```
TypeError: Cannot read property 'lineItems' of null
    at index.js:4954
```

The stack trace points inside `CheckoutResource.fetch`:
```js
    value: function fetch(id) {
      var _this2 = this;

      return this.graphQLClient.send(query$11, { id: id }).then(defaultResolver('node')).then(function (checkout) {
        return _this2.graphQLClient.fetchAllPages(checkout.lineItems, { pageSize: 250 }).then(function (lineItems) {
          checkout.attrs.lineItems = lineItems;

          return checkout;
        });
      });
    }
```

**To Reproduce**

The following shopify checkout ID appears to cause this error anytime we load it using the library:
`gid://shopify/Checkout/3947ae67053e5743c5b6d3d085110cc4?key=7a716e4b6af89fad8d036d432c06712e`
Our shopify domain is:
`watermark-resources.myshopify.com`

The code which causes the error is:
```ts
console.log('fetch', checkoutID)  // checkoutID is "Z2lkOi8vc2hvcGlmeS9DaGVja291dC8zOTQ3YWU2NzA1M2U1NzQzYzViNmQzZDA4NTExMGNjND9rZXk9N2E3MTZlNGI2YWY4OWZhZDhkMDM2ZDQzMmMwNjcxMmU="
const checkout: ICheckout = await client.checkout.fetch(checkoutID)
```

In my Chrome network tab the response payload appears to be empty.
image

**Expected behavior**
The library should return the checkout if it exists, or return `null` if it does not exist.

**Environment (please complete the following information):**
 - OS: OSX, Heroku w/ Ruby buildpack
 - chrome, firefox
 - SDK Version`v1.8.0`

**Additional context**

**Bug Report Checklist**
- [x] I have read and agree to the [CODE_OF_CONDUCT.md](https://github.com/Shopify/js-buy-sdk/blob/master/CODE_OF_CONDUCT.md)
- [x] I have read the [CONTRIBUTING.md](https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md) guidelines.
- [x] I have provided a detailed description of the bug, including code samples, and/or network data.
- [x] I have provided information about my development environment, including SDK version.

rebeccajfriedman:
If you try to fetch a checkout which doesn't exist, the `CheckoutResource.fetch` code throws an error. Since this is a valid case, we should handle it by returning the null checkout.

Fixes #563

### DIFF
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -34,6 +34,8 @@ class CheckoutResource extends Resource {
       .send(checkoutNodeQuery, {id})
       .then(defaultResolver('node'))
       .then((checkout) => {
+        if (!checkout) { return null; }
+
         return this.graphQLClient.fetchAllPages(checkout.lineItems, {pageSize: 250}).then((lineItems) => {
           checkout.attrs.lineItems = lineItems;
 
--- a/fixtures/node-null-fixture.js
+++ b/fixtures/node-null-fixture.js
@@ -1,0 +1,5 @@
+export default {
+  "data": {
+    "node": null
+  }
+};
--- a/test/client-checkout-integration-test.js
+++ b/test/client-checkout-integration-test.js
@@ -4,6 +4,7 @@ import fetchMock from './isomorphic-fetch-mock'; // eslint-disable-line import/n
 
 // fixtures
 import checkoutFixture from '../fixtures/checkout-fixture';
+import checkoutNullFixture from '../fixtures/node-null-fixture';
 import checkoutCreateFixture from '../fixtures/checkout-create-fixture';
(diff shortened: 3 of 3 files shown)


src/checkout-resource.js
```diff
@@ -34,6 +34,8 @@ class CheckoutResource extends Resource {
       .send(checkoutNodeQuery, {id})
       .then(defaultResolver('node'))
       .then((checkout) => {
+        if (checkout === null) { return null; }
```
minasmart: Does this need an explicit null check or would `!checkout` suffice?

PR #567, opened 2018-10-22
Expose discounts on checkouts

rebeccajfriedman:
This updates `schema.json` to include: articles, blogByHandle, blogs, discountAllocation, discountCodeApplication, manualDiscountApplication, scriptDiscountApplication, some other fields on the customer, checkout and order objects and some additional mutations.

Other changes:
- Add a mutation to remove an applied discount code
- Exposes discountApplications on the checkout object
- Exposes discountAllocations on the checkoutLineItem object
- Add a test for the addDiscount mutation

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,17 @@ client.checkout.addDiscount(checkoutId, discountCode).then(checkout => {
 });
 ```
 
+### Removing a Discount
+```javascript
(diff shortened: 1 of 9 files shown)


StefanSlehta:
I think we'll have to wait a bit before this can merge until Github webhooks issue is resolved. The last build for this branch was 5 days ago. Or we can trigger a build manually ?

PR #585, opened 2018-10-30
Use checkoutAttributesUpdateV2Mutation rather than checkoutAttributesUpdateMutation

jamiemtdwyer:
Update `CheckoutResource.updateAttributes()` to use the checkoutAttributesUpdateV2Mutation rather than the deprecated checkoutAttributesUpdateMutation.

This does not change the public API. I'll release a new patch version after this merges.

Closes https://github.com/Shopify/shopify/issues/176477.


rebeccajfriedman:
What about CheckoutEmailUpdateV2 and CheckoutDiscountCodeApplyV2 - are they coming in separate PRs?

jamiemtdwyer:
Good points @minasmart – I'll 🎩 to see how the behaviour changes for the client.

> What about CheckoutEmailUpdateV2 and CheckoutDiscountCodeApplyV2 - are they coming in separate PRs?

I'll update this PR to add those as well

jamiemtdwyer:
```
client.checkout.updateAttributes(checkoutId, {customAttributes: [{key: "MyKey", notValue: "MyValue"}]}).then((checkout) => {
  thisCheckout = checkout;
});
Promise {<pending>}__proto__: Promise[[PromiseStatus]]: "rejected"[[PromiseValue]]: TypeError: Cannot read property 'checkoutAttributesUpdateV2' of undefined
    at http://127.0.0.1:8080/dist/buybutton.js:3542:24
buybutton.js:3542 Uncaught (in promise) TypeError: Cannot read property 'checkoutAttributesUpdateV2' of undefined
    at buybutton.js:3542
```

@minasmart 🎩'ing this with buybutton.js – I think if I'm interpreting your comment correctly, then we won't need a major version bump here, since it still seems to trigger the catch branch in the Promise?

minasmart:
If the new (null checkout, but not a server error), and old (null checkout, but server error) both trigger the catch branch, then clients won't have to change how they are handling anything. That's great! 🚀

minasmart:
Striking my last comment, does the mutation with a null checkout not return some userErrors to inform people of why?

I thought we had something in here to reject the promise if there are user errors. https://github.com/Shopify/js-buy-sdk/blob/master/src/handle-checkout-mutation.js#L20

Shouldn't the mutation provide some context as to why the checkout is null?

jamiemtdwyer:
The result when calling `client.checkout.updateAttributes()` with a bad input seems to be the same both on this branch and the latest published version

this branch:
```
> const badInput = {customAttributes: [{key: ""}]};
undefined
> client.checkout.updateAttributes(checkoutId, badInput).then((checkout) => {
... myCheckout = checkout;
... });
Promise {
  <pending>,
  domain:
   Domain {
     domain: null,
     _events:
      { removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [] } }
> (node:50673) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'checkoutAttributesUpdateV2' of undefined
    at /Users/jamied/src/github.com/Shopify/js-buy-sdk/index.js:3573:24
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:50673) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:50673) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

1.9.1:
```
> client.checkout.updateAttributes(checkoutId, badInput).then((checkout) => {
... myCheckout = checkout;
... });
Promise {
  <pending>,
  domain:
   Domain {
     domain: null,
     _events:
      { removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [] } }
> (node:20282) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'checkoutAttributesUpdate' of undefined
    at /Users/jamied/node_modules/shopify-buy/index.js:3573:24
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:20282) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:20282) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The line being referenced in the stack trace is [here, in handle-checkout-mutation.js](https://github.com/Shopify/js-buy-sdk/blob/master/src/handle-checkout-mutation.js#L3). Given this, I think we can probably merge this without deploying and address this issue in follow-up PR

rebeccajfriedman:
I tested calling `client.checkout.updateAttributes` with an invalid Checkout ID on both this branch and the current master branch (v1.9.1).

v1.9.1
```
TypeError: Cannot read property 'checkoutAttributesUpdate' of undefined
    at /HOME/js-buy-sdk/index.js:3573:24
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

This branch:
```
Error: [{"field":["checkoutId"],"message":"Checkout does not exist"}]
    at /HOME/js-buy-sdk/index.js:3591:29
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

So as suspected, there is no breaking functionality. The only visible change is to the quality of the error message.

minasmart:
That sounds great! No need for a major bump then. Just a minor and 🚢

Issue #594, opened 2018-11-13
Field availableForSale not returned when fetching all products


philspitler:
When fetching all products using the js-buy-sdk, "client.product.fetchAll()", the **availableForSale** _Getter_ is no longer available.  This did work, but I don't see a code change in this library or this would be a pull request instead of an issue.  My code worked for quite some time with this working.  This seems like a very recent change.

**To Reproduce**
Call client.product.fetchAll() and see the results, they do not contain the **availableForSale** _Getter_.

Please see returned Objects below:
```javascript
GraphModel {
    id: [Getter],
    createdAt: [Getter],
    updatedAt: [Getter],
    descriptionHtml: [Getter],
    description: [Getter],
    handle: [Getter],
    productType: [Getter],
    title: [Getter],
    vendor: [Getter],
    tags: [Getter],
    publishedAt: [Getter],
    onlineStoreUrl: [Getter],
    options: [Getter],
    images: [Getter],
    variants: [Getter],
    refetchQuery: [Getter],
    type: [Getter],
    nextPageQueryAndPath: [Function],
    hasNextPage: Scalar { value: true },
    hasPreviousPage: Scalar { value: true },
    variableValues: { first: 20 } },
  GraphModel {
    id: [Getter],
    createdAt: [Getter],
    updatedAt: [Getter],
    descriptionHtml: [Getter],
    description: [Getter],
    handle: [Getter],
    productType: [Getter],
    title: [Getter],
    vendor: [Getter],
    tags: [Getter],
    publishedAt: [Getter],
    onlineStoreUrl: [Getter],
    options: [Getter],
    images: [Getter],
    variants: [Getter],
    refetchQuery: [Getter],
    type: [Getter],
    nextPageQueryAndPath: [Function],
    hasNextPage: Scalar { value: true },
    hasPreviousPage: Scalar { value: true },
```

philspitler:
Please see the documentation for the product object:

https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/object/product#fields

RE:
```
availableForSale   (Boolean!)
Indicates if at least one product variant is available for sale.
```

And this was working just a couple days ago.

jamiemtdwyer:
Sorry about that, I see that now. The linked PR will add `availableForSale` to the Product query fragment.

As for this working a couple days ago, I'm not sure how that would have been possible. The field is not included in the query fragment, and the history for that file does not show that it's been included at any point.

philspitler:
[...]

PR #595, opened 2018-11-13
Adds availableForSale to Product query fragment

StefanSlehta:
Addresses #594

### DIFF
--- a/fixtures/product-fixture.js
+++ b/fixtures/product-fixture.js
@@ -3,6 +3,7 @@ export default {
     "node": {
       "__typename": "Product",
       "id": "Z2lkOi8vc2hvcGlmeS9Qcm9kdWN0Lzc4NTc5ODkzODQ=",
+      "availableForSale": true,
       "createdAt": "2016-09-25T21:31:33Z",
       "updatedAt": "2017-03-29T15:25:32Z",
       "descriptionHtml": "send me this cat",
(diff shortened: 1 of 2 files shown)


jamiemtdwyer:
LGTM, but I think this will need to be accompanied by a major version bump unless we do some mapping on the client side? I'm not sure yet how this should be handled. I'm working on replacing some other deprecated fields in parallel on another branch currently.

src/graphql/VariantFragment.graphql
```diff
@@ -1,5 +1,6 @@
 fragment VariantFragment on ProductVariant {
   id
+  availableForSale
```
jamiemtdwyer: we should also remove `available` here since it's deprecated, and `availableForSale` replaces it

philspitler:
> LGTM, but I think this will need to be accompanied by a major version bump unless we do some mapping on the client side? I'm not sure yet how this should be handled. I'm working on replacing some other deprecated fields in parallel on another branch currently.

Hey @jamiemtdwyer - I don't understand how adding an extra data point to what is returned from the storefront API breaks anything.  If you are saying you are going to attach this to other breaking changes as an arbitrary dependency, I have a hard time making sense of that.  Will you please explain?

StefanSlehta:
Hey @philspitler ! I think @jamiemtdwyer was referring to removing the deprecated `available` field as a breaking change requiring a major version bump, and not the addition of `availableForSale`

Issue #586, opened 2018-11-01
Uses deprecated field names in GraphQL queries


freder:
<img width="370" alt="screenshot 2018-11-01 at 20 57 54" src="https://user-images.githubusercontent.com/724844/47876164-e9eced00-de18-11e8-9cbd-475d5021016e.png">

these are the ones I noticed. there might be more.

rebeccajfriedman:
Thank you for flagging this. These fields are deprecated, but we have not yet set a removal date. If you have time, I encourage you to create a PR to include the new fields that have replaced these.

BitterChris:
We've added aliasing for `available` and `src` to their new field names of `availableForSale` and `originalSrc` respectively in #600. This allows us to keep this library consistent for now, while querying the new non-deprecated fields from the Storefront API.

PR #596, opened 2018-11-14
[DNM] Replace deprecated fields with their non-deprecated counterparts

jamiemtdwyer:
Closes https://github.com/Shopify/shopify/issues/178821.

Updates the SDK to remove the use of deprecated fields and replace them with their non-deprecated counterparts.

* Replace `countryCode` on MailingAddress with `countryCodeV2`
* Replace `src` on Image with `originalSrc`
* Replace `available` on Productvariant with `availableForSale`

### DIFF
--- a/fixtures/checkout-create-fixture.js
+++ b/fixtures/checkout-create-fixture.js
@@ -54,7 +54,7 @@ export default {
           "province": "Ontario",
           "zip": "M3O 0W1",
           "name": "Meow Meowington",
-          "countryCode": "CA",
+          "countryCodeV2": "CA",
           "provinceCode": "ON",
           "id": "Z2lkOi8vc2hvcGlmeS9QcmdfnAU8nakdWMnAbh890hyOTEwNjA2NDU4NA=="
         },
--- a/fixtures/checkout-discount-code-apply-fixture.js
+++ b/fixtures/checkout-discount-code-apply-fixture.js
@@ -40,7 +40,7 @@ export default {
           "province": "Ontario",
           "zip": "M3O 0W1",
           "name": "Meow Meowington",
-          "countryCode": "CA",
+          "countryCodeV2": "CA",
           "provinceCode": "ON",
           "id": "291dC9lM2JkNzHJnnf8a89njNJNKAhu1gn7lMzM5MzFlZj9rZXk9NDcwOTJ=="
         },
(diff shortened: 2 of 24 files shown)


jamiemtdwyer:
Closes https://github.com/Shopify/js-buy-sdk/issues/586.

swalkinshaw:
Note that this is a breaking change. We can either be fine with that and use sem ver to properly bump the version, or we can use field aliases to keep the field names the same. But that only works if the return types remain the same.

jamiemtdwyer:
Yup, for sure I agree with the major version bump.

I'm moving this back to [WIP] since it looks like we might need to handle selectively fetching `orignalSrc` or `transformedSrc` fields, since the SDK is currently exercising the `src` field to perform image transformations.

jamiemtdwyer:
Yeah, the new return values are typed differently. The only one that remains the same is the image `src` -> `originalSrc`. I think it might be somewhat reasonable to add an alias here, but I'm not super sure how to accomplish that. Has it been done here before?

minasmart:
They're the same types in JS everywhere as far as I can tell. This would be a breaking change if we were instantiating typed models, but the JS primitives are all the same. Am I missing something?

Also, aliases _should_ work with the underlying library.

ex:

```graphql
... on Product {
  src: originalSrc
}
```

jamiemtdwyer:
> This would be a breaking change if we were instantiating typed models, but the JS primitives are all the same.

Good point – I'll see what I can do. I'm in favour of breaking the public API as little as possible. 👍

rebeccajfriedman:
This is my thinking:
One of our goals for the future is figuring out how to make it easy to use custom fragments along with the SDK. In that scenario, we'd probably want to avoid aliasing fields so that the response from a custom fragment vs from a default fragment have the same fields.

So what I think we should do:
We should determine a cadence for breaking changes (say, once every 3 months). In between breaking changes, we should alias fields (so at least we can remove deprecated fields from being queried). When we release a breaking change, we should remove the aliases.

IMO we can use this opportunity as our first breaking change release.

rebeccajfriedman:
@jamiemtdwyer discussed offline further with Melissa and Mina and this is my proposal: Let's release as a non-breaking change on November 27 and then schedule a breaking change for the new year (possibly December).

Can you please update this PR?

jamiemtdwyer:
Closed in favour of https://github.com/Shopify/js-buy-sdk/pull/600/.

Issue #597, opened 2018-11-14
Uncaught (in promise) when calling client.product.fetchAll()


baptistebriel:
**Describe the bug**
I cannot fetchAll products from a simple `buildClient`, the exact same code as showned in the getting started demos.

<img width="818" alt="screen shot 2018-11-14 at 2 52 40 pm" src="https://user-images.githubusercontent.com/2464663/48517993-fd3a8800-e81c-11e8-8401-e6039b520042.png">

**To Reproduce**
```javascript
const client = Client.buildClient({
  domain: 'subdomain.myshopify.com',
  storefrontAccessToken: 'access-token'
});

client.product.fetchAll().then((products) => {
  // Do something with the products
  console.log(products);
});
```

**Expected behavior**
Should get a JS object list of all the products listed on the Shopify store.

**Environment (please complete the following information):**
 - All
 - Tested on Chrome
 - SDK Version 1.10.0

**Additional context**

Weird thing is, if I try to access the .json file using `https://user:pass@subdomain.myshopify.com/admin/products.json` from the StoreFront API, I do get a list of products. Seems like the issue I'm having is only with the Shopify Buy SDK.
Is there something I'm missing here? Thanks a lot for the help.

**Bug Report Checklist**
- [x] I have read and agree to the [CODE_OF_CONDUCT.md](https://github.com/Shopify/js-buy-sdk/blob/master/CODE_OF_CONDUCT.md)
- [x] I have read the [CONTRIBUTING.md](https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md) guidelines.
- [x] I have provided a detailed description of the bug, including code samples, and/or network data.
- [x] I have provided information about my development environment, including SDK version.

mikkoh:
@rebeccajfriedman I looked into this issue. I'm able to reproduce this on my test shop if I disable the `Read Product Tags` permission.

<img width="598" alt="screen shot 2018-11-15 at 4 33 58 pm" src="https://user-images.githubusercontent.com/496903/48583224-8d97cc00-e8f4-11e8-8f87-4a8654882e0d.png">

@baptistebriel try it out. See if checking/allowing this permission helps. You can temporarily enable it. If there's a reason why you don't want this permission enabled your best option might be to wait for a fix.

baptistebriel:
That fixed the access denied issue! Thanks a lot!

igghera:
[...]

PR #598, opened 2018-11-17
Remove tags field from Product fragment #597

igghera:
The default ProductFragment provided by this SDK comes with a `tags` field,
but since the permission to read product tags is disabled by default in the
Storefront API settings, an error would occur with the message "access denied"
when using `client.product.fetchAll()`.

This PR fixes the issue by removing the `tags` field from the Product fragment.

### DIFF
--- a/fixtures/collection-with-products-fixture.js
+++ b/fixtures/collection-with-products-fixture.js
@@ -27,9 +27,6 @@ export default {
               "productType": "cat",
               "title": "Cat",
               "vendor": "sendmecats",
-              "tags": [
-                "vintage"
-              ],
               "publishedAt": "2017-01-12T19:44:42Z",
               "options": [
                 {
@@ -185,9 +182,6 @@ export default {
               "productType": "cat",
               "title": "Cat 2",
               "vendor": "sendmecats",
-              "tags": [
-                "fluffy"
-              ],
               "publishedAt": "2017-02-03T18:52:27Z",
               "options": [
                 {
--- a/fixtures/product-by-handle-fixture.js
+++ b/fixtures/product-by-handle-fixture.js
@@ -11,11 +11,6 @@ export default {
         "handle": "cape-dress-1",
         "productType": "women's dresses",
         "vendor": "Co",
-        "tags": [
-          "arrivals",
-          "AW15",
-          "black",
-        ],
         "publishedAt": "2017-02-15T15:09:45Z",
         "options": [
           {
--- a/fixtures/product-fixture.js
+++ b/fixtures/product-fixture.js
@@ -11,9 +11,6 @@ export default {
       "productType": "cat",
       "title": "Cat",
       "vendor": "sendmecats",
-      "tags": [
-        "vintage"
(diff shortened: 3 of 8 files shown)


rebeccajfriedman:
Closes https://github.com/Shopify/js-buy-sdk/issues/597

rebeccajfriedman:
Tests are failing here because the fixtures are still returning `tags`, even though there is no longer a `tags` field being sent in the queries. You'll need to remove `tags` from all the fixtures to get it to pass.

gilesparallel:
I want the tags to be returned with the product resource. Is there any way to re-enable this?

priyankafet:
I would need tags in response as this is breaking whole site.

rebeccajfriedman:
@priyankafet @gilesparallel This change was released in version `1.11.0`. In the immediate term, you can just lock to version `1.10.0` and you will have tags back. We are working on a way to allow more flexibility with the queries, but you can also patch the SDK by writing your own product fragment and passing it through your own method.

petomi:
Any news on when we'll be able to get tags returned again? It's essential for our site's functionality.

AlejoYarce:
@rebeccajfriedman I want to know more about this **...but you can also patch the SDK by writing your own product fragment and passing it through your own method.**

We need to use the tags but don't want to have an old version.

petomi:
We're in the same boat, unfortunately. We need tags and using an old version is ok for now, but we need a more permanent solution. I've tried updating the product fragment definition in the source code, but I can't get the project to build properly (even pulling directly from master it doesn't build).
Not sure if I'm missing something...

rebeccajfriedman:
I [opened a PR](https://github.com/Shopify/js-buy-sdk/pull/611) to start publishing a version of the SDK with a full Storefront API GraphQL schema exposed. This will make it easier to patch the SDK. Once it's lived, I'll update the repo README.md to provide instructions on how you can use it. I'm hoping to release the PR next week, and I'll post here when I do.

rebeccajfriedman:
Please read the section in the README entitled [Expanding the SDK](https://github.com/Shopify/js-buy-sdk#expanding-the-sdk) for more information on how you can write your own queries using the unoptimized umd build.

AlejoYarce:
@rebeccajfriedman 
```
 root.addConnection('products', {args: {first: 10}}, (product) => {
    product.add('title');
    product.add('tags');// Add fields to be returned
  });
```

If I want to get **Variants** with some fields, how I can add that field? Just adding **product.add('variants')** or there is another way to add the Variants "fragment"

rebeccajfriedman:
For any field that is defined as a connection in the [documentation](https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/object/product), you can use `addConnection`.

```
root.addConnection('products', {args: {first: 10}}, (product) => {
    product.add('title');
    product.add('tags');// Add fields to be returned
    product.addConnection('variants', {args: {first: 1}}, (variant) => {
        variant.add('id');
    });
  });
```

PR #606, opened 2018-12-24
Bump Storefront API Schema

rebeccajfriedman:
Key Changes:
- Remove unused fixtures
- Fetch `collections`, `products`, `collectionByHandle` and `productByHandle` from the QueryRoot instead of from the Shop object.
- Return `checkoutUserErrors` instead of `userErrors` when present. This will add an extra field called `code`.
- Expose `checkoutLineItemsReplace` mutation (to replace deprecated `checkoutLineItemsAdd `, `checkoutLineItemsRemove `, and `checkoutLineItemsUpdate ` mutations).


rebeccajfriedman:
Yes, I think so. Favouring `checkoutUserErrors` over `userErrors` could be a breaking change to API consumers who are not currently handling `userErrors`.

PR #611, opened 2019-01-11
Create unoptimized packages as well

rebeccajfriedman:
Looking for feedback on:
1) I couldn't figure out how to have an array of exports from `rollup.config.js` on our current rollup version, so I duped the file. Is there a better way to do this?
2) I'm only generating an unoptimized version of the umd file (and a minified one). Is this sufficient or should I regenerate them all?
3) This was my first time really working with `package.json`. Can the scripts be rewritten to be less competitive?
4) How should I test this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "scripts": {
     "prepublish": "yarn run build",
     "build": "yarn run build:optimized && yarn run build:unoptimized",
-    "build:optimized": "rollup -c && yarn run minify-umd",
+    "build:optimized": "rollup -c && yarn run minify-umd:optimized",
     "build:unoptimized": "rollup -c rollup-unoptimized.config.js && yarn run minify-umd:unoptimized",
     "start": "yarn run clean && yarn run setup-paths && concurrently --names DEPS,LIVERELOAD,ROLLUP,HTTP,NOTIFIER 'yarn run test:web-deps' 'yarn run test:livereload-server' 'yarn run test:watch' 'http-server -p 4200 .tmp/test' 'yarn run print-start-message'",
     "clean": "rimraf .tmp index.*",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -11,7 +11,9 @@
   },
   "scripts": {
     "prepublish": "yarn run build",
-    "build": "rollup -c && yarn run minify-umd",
+    "build": "yarn run build:optimized && yarn run build:unoptimized",
```
minasmart: this can be `concurrently`, since one doesn't depend on the other

package.json
```diff
@@ -27,7 +29,8 @@
     "lint:reporter-args": "test -n \"${CI}\" && >&2 echo -o $CIRCLE_TEST_REPORTS/junit/eslint.xml -f junit",
     "print-start-message": "wait-on file:.tmp/test/index.html file:.tmp/test/tests.js tcp:35729 tcp:4200 && echo \"\n\n⚡️⚡️⚡️  Good to go at http://localhost:4200 ⚡️⚡️⚡️\"",
     "schema:fetch": "graphql-js-schema-fetch --url 'https://graphql.myshopify.com/api/graphql' --header 'Authorization: Basic MzUxYzEyMjAxN2QwZjJhOTU3ZDMyYWU3MjhhZDc0OWM=' | jq '.' > schema.json",
-    "minify-umd": "echo \"/*\n$(cat LICENSE.txt)\n*/\" > index.umd.min.js && babel-minify index.umd.js >> index.umd.min.js"
+    "minify-umd": "echo \"/*\n$(cat LICENSE.txt)\n*/\" > index.umd.min.js && babel-minify index.umd.js >> index.umd.min.js",
```
minasmart: should probably be explicit with names and use `optimized` and `unoptimized`

rollup-unoptimized.config.js
```diff
@@ -0,0 +1,22 @@
...
+
+baseConfig.plugins.unshift(
+  graphqlCompiler({
+    schema: './schema.json',
+    optimize: false,
+    profileDocuments: ['src/graphql/**/*.graphql']
+  })
+);
+
+baseConfig.targets = [
```
minasmart: since this is just reassigned, `targets` should probably not be part of base and should be promoted to the consumed config.

package.json
rebeccajfriedman: I tried to use `concurrently 'yarn run build:optimized' 'yarn run build:unoptimized'`, but it threw up in the minify step. Maybe it was something with the `cat`? Or did I use it incorrectly?

minasmart: Accidental resolve. I think this happens because cat calls fflush on the license file in parallel which won’t work. You may want to prepend the license in a separate step using cat and piping to tee with the prepend option, then concurrently minify

rebeccajfriedman: I'm going to skip this for now, but I'll keep in mind that it can be refactored better!

PR #613, opened 2019-01-14
Update README with unoptimized version info

rebeccajfriedman:
Updates the README with an example on how to mod the JS Buy SDK queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,8 @@ import Client from 'shopify-buy/index.amd';
 import Client from 'shopify-buy/index.umd';
 ```
 **UMD Unoptimized:**
+This will be larger than the optimized version, as it will contain all fields that are available in the [Storefront API](https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference). This should only be used when you need to add custom queries to supplement the JS Buy SDK queries.
+
 ```javascript
 import Client from 'shopify-buy/index.unoptimized.umd';
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -215,6 +229,38 @@ client.checkout.removeDiscount(checkoutId).then(checkout => {
 });
 ```
 
+## Expanding the SDK
+
+Not all fields that are available on the [Storefront API](https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference) are exposed through the SDK. If you use the unoptimized version of the SDK, you can easily build your own queries. In order to do this, use the UMD Unoptimized build.
+
+### Initializing the Client
+```javascript
+import Client from 'shopify-buy/umd';
```
rebeccajfriedman: How can I find out what this should be?

README.md
```diff
@@ -73,6 +83,10 @@ import Client from 'shopify-buy/index.amd';
 ```javascript
 import Client from 'shopify-buy/index.umd';
 ```
+**UMD Unoptimized:**
```
minasmart: You should maybe include a definition of what unoptimized means. The term is very inside baseball as it is.

minasmart: At least in terms of the cases where this version is useful.

README.md
```diff
@@ -215,6 +229,38 @@ client.checkout.removeDiscount(checkoutId).then(checkout => {
 });
 ```
 
+## Expanding the SDK
+
+Not all fields that are available on the [Storefront API](https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference) are exposed through the SDK. If you use the unoptimized version of the SDK, you can easily build your own queries. In order to do this, use the UMD Unoptimized build.
+
+### Initializing the Client
```
minasmart: This should specify which client. By default, we want to recommend the optimized client, but for special cases, use the thicker one.

minasmart: (I know it's under the expanding the sdk section. But for the sake of people skimming or looking at just the headings it'd be useful)

PR #631, opened 2019-02-13
Add checkoutUserError field selection to remaining checkout mutations.

jamiemtdwyer:
Adds checkoutUserError field selection to remaining checkout mutations that were lacking them. This adds consumable error codes to all of the checkout mutation payloads.

Also add tests for the new fields, and add tests to some other mutations that were lacking them.

Fixes an issue with `checkoutLineItemsReplace` mutation where userErrors (which is of type CheckoutUserErrors in the schema) was not using the correct query fragment.

### DIFF
--- a/fixtures/checkout-shipping-address-update-v2-with-user-errors-fixture.js
+++ b/fixtures/checkout-shipping-address-update-v2-with-user-errors-fixture.js
@@ -4,9 +4,17 @@ export default {
       "userErrors": [
         {
           "message": 'Country is not supported',
-          "field": ['shippingAddress country']
+          "field": ['shippingAddress', 'country'],
+        },
+      ],
+      "checkoutUserErrors" : [
+        {
+          "message": 'Country is not supported',
+          "field": ['shippingAddress', 'country'],
+          "code": 'NOT_SUPPORTED'
         }
-      ]
+      ],
+      "checkout": null
     }
   }
 };
--- a/src/graphql/checkoutLineItemsRemoveMutation.graphql
+++ b/src/graphql/checkoutLineItemsRemoveMutation.graphql
@@ -3,6 +3,9 @@ mutation ($checkoutId: ID!, $lineItemIds: [ID!]!) {
     userErrors {
       ...UserErrorFragment
     }
+    checkoutUserErrors {
+      ...CheckoutUserErrorFragment
+    }
     checkout {
       ...CheckoutFragment
     }
--- a/src/graphql/checkoutLineItemsReplaceMutation.graphql
+++ b/src/graphql/checkoutLineItemsReplaceMutation.graphql
@@ -1,7 +1,7 @@
 mutation ($checkoutId: ID!, $lineItems: [CheckoutLineItemInput!]!) {
   checkoutLineItemsReplace(checkoutId: $checkoutId, lineItems: $lineItems) {
     userErrors {
-      ...UserErrorFragment
+      ...CheckoutUserErrorFragment
     }
     checkout {
       ...CheckoutFragment
--- a/src/graphql/checkoutShippingAddressUpdateV2Mutation.graphql
+++ b/src/graphql/checkoutShippingAddressUpdateV2Mutation.graphql
@@ -3,6 +3,9 @@ mutation checkoutShippingAddressUpdateV2($shippingAddress: MailingAddressInput!,
     userErrors {
       ...UserErrorFragment
     }
+    checkoutUserErrors {
+      ...CheckoutUserErrorFragment
+    }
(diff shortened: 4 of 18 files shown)


src/graphql/checkoutLineItemsReplaceMutation.graphql
```diff
@@ -1,7 +1,7 @@
 mutation ($checkoutId: ID!, $lineItems: [CheckoutLineItemInput!]!) {
   checkoutLineItemsReplace(checkoutId: $checkoutId, lineItems: $lineItems) {
     userErrors {
-      ...UserErrorFragment
+      ...CheckoutUserErrorFragment
```
spencercanner: Does this not have the checkoutUserErrors field because it's newer than the other line item mutations?

test/client-checkout-integration-test.js
```diff
@@ -112,7 +118,22 @@ suite('client-checkout-integration-test', () => {
     });
   });
 
-  test('it resolves with a checkout on Client.checkout#email_update', () => {
+  test('it resolve with user errors on Client.checkout#updateAttributes when input is invalid', () => {
```
spencercanner: Small nitpick on this test and a few below - `it resolves ...`

test/client-checkout-integration-test.js
```diff
@@ -175,6 +238,25 @@ suite('client-checkout-integration-test', () => {
     });
   });
 
+  test('it resolves with user errors on Client.checkout#updateLineItems when variant is invalid', () => {
+    fetchMock.postOnce(apiUrl, checkoutLineItemsUpdateWithUserErrorsFixture);
```
spencercanner: For consistency with other tests, could we move this line below the `const` declarations? Same for the test below

src/graphql/checkoutLineItemsRemoveMutation.graphql
```diff
@@ -3,6 +3,9 @@ mutation ($checkoutId: ID!, $lineItemIds: [ID!]!) {
     userErrors {
       ...UserErrorFragment
     }
+    checkoutUserErrors {
```
spencercanner: Does `checkout-line-items-remove-fixture` need an empty array for this field?

src/graphql/checkoutShippingAddressUpdateV2Mutation.graphql
```diff
@@ -3,6 +3,9 @@ mutation checkoutShippingAddressUpdateV2($shippingAddress: MailingAddressInput!,
     userErrors {
       ...UserErrorFragment
     }
+    checkoutUserErrors {
```
spencercanner: Does the `checkout-shipping-address-update-v2-fixture` need an empty array added for this field?

fixtures/checkout-shipping-address-update-v2-with-user-errors-fixture.js
```diff
@@ -4,7 +4,14 @@ export default {
...
           "message": 'Country is not supported',
-          "field": ['shippingAddress country']
+          "field": ['shippingAddress', 'country'],
+        },
+      ],
+      "checkoutUserErrors" : [
+        {
+          "message": 'Country is not supported',
+          "field": ['shippingAddress', 'country'],
+          "code": 'NOT_SUPPORTED'
```
spencercanner: Does this mutation also return a `null` checkout like the others?

src/graphql/checkoutLineItemsReplaceMutation.graphql
jamiemtdwyer: Yeah, right on. This mutation is newer so it doesn't have the legacy `userErrors`, so we don't need to distinguish between the legacy `userErrors` (of type [UserError!]!) and the newer checkout user errors type ([CheckoutUserError!]!)

jamiemtdwyer: https://help.shopify.com/en/api/custom-storefronts/storefront-api/reference/mutation/checkoutlineitemsreplace

src/graphql/checkoutShippingAddressUpdateV2Mutation.graphql
jamiemtdwyer: Good catch

jamiemtdwyer:
Addressed @spencercanner's comments in the latest commit

src/graphql/checkoutLineItemsReplaceMutation.graphql
rebeccajfriedman: Should we leave a note about this so that someone doesn't think it's a typo and change it?

Issue #614, opened 2019-01-15
No type of Automatic Discount Application found in schema


donovanroubos:
**Describe the bug**
On our webshop we've automatic discounts. The user wasn't able to add more than 3 products because the API gave an error: "No type of Automatic Discount Application found in schema".

**Environment**
 - OS: iOS
 - Browser: Safari, Chrome and Firefox
 - SDK Version `v1.9.1`

**Additional context**
The symatic versioning of the package had a breaking change in `v1.9.1`. For this version you had to communicate that it was broken and for the future it's better to do a bigger version bump. So that it wouldn't auto update on all applications.

We've fixed the error by downgraded the package version from `v^1.8.0` satisfied `v1.9.1` to `v1.8.0`. This isn't the expected behavior of such a big api.

rebeccajfriedman:
Thank you for flagging this. We are tracking this issue in our backend. Automatic Discounts are supported in the latest version (`v2.0.1`). Sorry for any inconvenience that this has caused you.

rebeccajfriedman:
This is foward fixed in the latest release: [v2.1.1](https://github.com/Shopify/js-buy-sdk/releases/tag/v2.1.1). We have added support for falling back when interface objects can't be translated.

PR #638, opened 2019-03-06
Bump graphql-js-client to 0.11.0

rebeccajfriedman:
Forward fix for https://github.com/Shopify/js-buy-sdk/issues/614 as it includes https://github.com/Shopify/graphql-js-client/pull/121

### DIFF
--- a/test/fetch-mock-helper.js
+++ b/test/fetch-mock-helper.js
@@ -1,0 +1,12 @@
+export function fixtureWithHeaders(fixture) {
+  return {
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: fixture
+  };
+}
+
+export default function fetchMockPostOnce(fetchMock, apiUrl, fixture) {
+  fetchMock.postOnce(apiUrl, fixtureWithHeaders(fixture));
+}
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "expect.js": "0.3.1",
     "fetch-mock": "5.12.2",
     "fs-extra": "1.0.0",
-    "graphql-js-client": "0.9.2",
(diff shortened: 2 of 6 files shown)


test/fetch-mock-helper.js
```diff
@@ -0,0 +1,8 @@
+export default function fixtureWithHeaders(fixture) {
```
melissaluu: Do we need a test to test this helper function?

minasmart: That doesn't seem like a bad idea. Technically, if this breaks, it could invalidate a whole whack of tests. I think tho, with the more recent change to the client (ignoring headers and trying to parse json and falling back to text), this method will probably become unnecessary. @rebeccajfriedman, is that correct?

rebeccajfriedman: @minasmart you are right. We are going to hold off on merging that PR in `graphql-js-client` because we aren't so happy with the resultant code, and we were struggling to come up with a better approach.

test/fetch-mock-helper.js
```diff
@@ -0,0 +1,12 @@
...
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: fixture
+  };
+}
+
+export default function fetchMockPostOnce(fetchMock, apiUrl, fixture) {
+  fetchMock.postOnce(apiUrl, fixtureWithHeaders(fixture));
+}
```
m-ux: I'm curious as to what this wrapper is helpful for (I may be missing some context).  Is this required due to bumping the graph package, or is it just a syntactically preferred way to write these tests?

rebeccajfriedman: Discussed offline :)

Issue #636, opened 2019-02-26
ProductHelper variantForOptions documentation example is incorrect


bradgarropy:
**Describe the bug**
The `variantForOptions` example on the [ProductHelpers][1] page is incorrect.

The current example is as follows:
```
const selectedVariant = client.product.variantForOptions(product, {
  size: "Small",
  color: "Red"
});
```

**To Reproduce**
Navigate to the [ProductHelpers][1] page of the documentation.

**Expected behavior**
I believe the `variantForOptions` example should be as follows:
```
const selectedVariant = client.product.helpers.variantForOptions(product, {
  size: "Small",
  color: "Red"
});
```

**Environment (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - SDK Version: `v2.1.0`

**Additional context**
By following the example, I see an error.
```
Uncaught TypeError: client.product.variantForOptions is not a function
    at getVariant (shopify.js:56)
    at sleeves.js:44
    at commitHookEffectList (react-dom.development.js:17079)
    at commitPassiveHookEffects (react-dom.development.js:17103)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at commitPassiveEffects (react-dom.development.js:18570)
    at wrapped (scheduler-tracing.development.js:207)
    at flushFirstCallback (scheduler.development.js:107)
    at flushWork (scheduler.development.js:219)
    at MessagePort../node_modules/scheduler/cjs/scheduler.development.js.channel.port1.onmessage (scheduler.development.js:611)
```

By using `client.product.helpers.variantForOptions(product, options)`, everything works as expected.

I know the [CONTRIBUTING][2] guidelines say not to update the documentation, but I'm going to prepare a pull request for this minor fix anyways, so that you can merge it right away if you want to.

**Bug Report Checklist**
- [x] I have read and agree to the [CODE_OF_CONDUCT.md](https://github.com/Shopify/js-buy-sdk/blob/master/CODE_OF_CONDUCT.md)
- [x] I have read the [CONTRIBUTING.md](https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md) guidelines.
- [x] I have provided a detailed description of the bug, including code samples, and/or network data.
- [x] I have provided information about my development environment, including SDK version.

[...]

PR #639, opened 2019-03-06
Add correct documentation for variantForOptions from product helpers

jamiemtdwyer:
Updates the documentation decorator to use the correct call to the product resource helpers – it should be `client.product.helpers.variantForOptions`, rather than `client.product.variantForOptions` as `variantForOptions` is made available in the product resource through the `helpers` object.

https://github.com/Shopify/js-buy-sdk/blob/master/src/product-resource.js#L16-L19


jamiemtdwyer:
Closes #636.

PR #645, opened 2019-03-20
Bump schema

rebeccajfriedman:
PR to bump schema. Schema changes include:
- Adds currency support for Bermudian Dollar
- Adds `seo` field to `article` which exposes SEO title and description information
- Deprecates `status` field on `transaction` and replaces it with `statusV2`, which supports a null status. 
- Add line items subtotal price as a field returned on checkout.


rebeccajfriedman:
@jmignac Test failures look legit.

PR #667, opened 2019-05-07
Gift Cards addition and removal

agos:
Greetings,

this PR adds support for `Client.checkout#addGiftCards`, which allows clients to add gift cards to a checkout using their codes, and `Client.checkout#removeGiftCard`, which allows clients to remove a previously added gift card from a checkout using the addition id. This also adds the applied gift cards to `checkoutFragment`. This is meant as a natural companion to the apply/remove discount code methods.

Basic success / failure integration tests and their fixtures are also included.

I've left three separated commits: one for `addGiftCards`, one for the added field on the checkout fragment, one for `removeGiftCard`. This is to make it easy to separate them in case you prefer different pull request, but can obviously be squashed if it's not needed.

### DIFF
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -12,6 +12,7 @@ import checkoutLineItemsUpdateMutation from './graphql/checkoutLineItemsUpdateMu
 import checkoutAttributesUpdateV2Mutation from './graphql/checkoutAttributesUpdateV2Mutation.graphql';
 import checkoutDiscountCodeApplyV2Mutation from './graphql/checkoutDiscountCodeApplyV2Mutation.graphql';
 import checkoutDiscountCodeRemoveMutation from './graphql/checkoutDiscountCodeRemoveMutation.graphql';
+import checkoutGiftCardsAppendMutation from './graphql/checkoutGiftCardsAppendMutation.graphql';
 import checkoutEmailUpdateV2Mutation from './graphql/checkoutEmailUpdateV2Mutation.graphql';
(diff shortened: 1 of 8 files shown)


agos:
I have signed the CLA, but am unsure on how to re-run the CLA check!

src/checkout-resource.js
```diff
@@ -181,6 +183,48 @@ class CheckoutResource extends Resource {
...
+      .send(checkoutGiftCardsAppendMutation, {checkoutId, giftCardCodes})
+      .then(handleCheckoutMutation('checkoutGiftCardsAppend', this.graphQLClient));
+  }
+
+  /**
+   * Remove a gift card from an existing checkout
+   *
+   * @example
+   * const checkoutId = 'Z2lkOi8vc2hvcGlmeS9DaGVja291dC9kMTZmM2EzMDM4Yjc4N=';
+   * const giftCardId = 'Z2lkOi8vc2hvcGlmeS9BcHBsaWVkR2lmdENhcmQvNDI4NTQ1ODAzMTI=';
```
rebeccajfriedman: this should be named `appliedGiftCardId`

src/checkout-resource.js
```diff
@@ -181,6 +183,48 @@ class CheckoutResource extends Resource {
...
+   *
+   * client.checkout.removeGiftCard(checkoutId, appliedGiftCardId).then((checkout) => {
+   *   // Do something with the updated checkout
+   * });
+   *
+   * @param {String} checkoutId The ID of the checkout to add gift cards to.
+   * @param {String} giftCardId The gift card id to remove from the checkout.
```
rebeccajfriedman: Same as above, should be named `appliedGiftCardId `

PR #679, opened 2019-06-18
Add multicurrency fields to queries

jmignac:
Add multicurrency fields and connections to queries:

AppliedGiftCard: `presentmentAmountUsed`.
Checkout: `paymentDueV2`, `totalTaxV2`, `subtotalPriceV2`, `totalPriceV2`.
Order: `subtotalPriceV2`, `totalShippingPriceV2`, `totalTaxV2`, `totalPriceV2`, `totalRefundedV2`.
Variant: `presentmentPrices`.
Shop: `paymentSettings` (that includes `enabledPresentmentCurrencies`)

### DIFF
--- a/src/graphql/VariantFragment.graphql
+++ b/src/graphql/VariantFragment.graphql
@@ -2,6 +2,20 @@ fragment VariantFragment on ProductVariant {
   id
   title
   price
+  presentmentPrices(first: 20) {
+    pageInfo {
+      hasNextPage
+      hasPreviousPage
+    }
+    edges {
+      node {
+        price {
+          amount
+          currencyCode
+        }
+      }
+    }
+  }
   weight
   available: availableForSale
   sku
--- a/src/graphql/shopQuery.graphql
+++ b/src/graphql/shopQuery.graphql
@@ -1,6 +1,9 @@
 query {
   shop {
     currencyCode
+    paymentSettings {
+      enabledPresentmentCurrencies
+    }
     description
     moneyFormat
     name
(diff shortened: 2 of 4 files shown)


src/graphql/ProductFragment.graphql
```diff
@@ -7,6 +7,30 @@ fragment ProductFragment on Product {
   description
   handle
   productType
+  priceRange {
```
rebeccajfriedman: I don't think we need to include this. Let's only include functionality that we think most folks will use.

src/graphql/ProductFragment.graphql
```diff
@@ -7,6 +7,30 @@ fragment ProductFragment on Product {
...
+    minVariantPrice {
+      amount
+      currencyCode
+    }
+    maxVariantPrice {
+      amount
+      currencyCode
+    }
+  }
+  presentmentPriceRanges(first: 250) {
```
rebeccajfriedman: Same as above.

src/graphql/VariantFragment.graphql
```diff
@@ -2,6 +2,16 @@ fragment VariantFragment on ProductVariant {
   id
   title
   price
+  presentmentPrices(first: 250) {
```
rebeccajfriedman: Maybe first 20 is okay?

src/graphql/shopQuery.graphql
```diff
@@ -1,6 +1,15 @@
 query {
   shop {
     currencyCode
+    paymentSettings {
```
rebeccajfriedman: The only payment setting I would include is `enabledPresentmentCurrencies`. The other ones are not critical for a flow that uses web checkout.

jmignac:
> You added some V2 fields, but can you log a ticket somewhere to remove the V1 fields (we should do this later in the year)?

https://github.com/Shopify/shopify/issues/204399 👍

PR #691, opened 2019-08-07
Add priceV2 and compareAtPriceV2 to variant fragment

spencercanner:
Add the variant's `MoneyV2` fields (`priceV2` & `compareAtPriceV2`) to the variant fragment so users can move away from using the deprecated `price` and `compareAtPrice` fields

### DIFF
--- a/fixtures/product-by-handle-fixture.js
+++ b/fixtures/product-by-handle-fixture.js
@@ -107,6 +107,14 @@ export default {
               "available": true,
               "weight": 0,
               "price": "788.00",
+              "priceV2": {
+                "amount": "788.00",
+                "currencyCode": "CAD"
+              },
+              "compareAtPriceV2": {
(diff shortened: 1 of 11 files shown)


rebeccajfriedman:
Might be worth adding those fields to the fixtures too!

fixtures/product-by-handle-fixture.js
```diff
@@ -127,6 +135,14 @@ export default {
               "available": true,
               "weight": 0,
               "price": "788.00",
+              "priceV2": {
+                "amount": "0.00",
```
rebeccajfriedman: It's a fixture, so it doesn't really matter that the priceV2 is being set to 0 here, but just calling out that the variant price is 788.

spencercanner: Fixed!

PR #694, opened 2019-08-12
Add custom source header

melissaluu:
Add custom `X-SDK-Variant-Source` header.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -47,6 +47,10 @@ class Client {
       'X-Shopify-Storefront-Access-Token': config.storefrontAccessToken
     };
 
+    if (config.source) {
+      headers['X-SDK-Variant-Source'] = config.source;
+    }
+
     if (fetchFunction) {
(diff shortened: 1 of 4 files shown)


minasmart:
Should this option be documented in the readme?

melissaluu:
Since this "feature" won't benefit 3rd party developers, I don't think we need to put it in the README. Also, since this value will represent the user of JS Buy SDK, constraining the possible values within JS Buy SDK will require direct the library to know its possible users.

minasmart:
> constraining the possible values within JS Buy SDK will require direct the library to know its possible users.

If it's only for internal tracking, then we do know the possible users. My concern is that this starts to get a ton of weird values in data land and people start using it in unexpected ways. If this isn't a possibility or concern though, then 👍.

rebeccajfriedman:
I think we're good to try this out as is, and see how it goes! 👍

PR #717, opened 2019-10-16
Add an optional productsFirst param to fetchWithProducts

### LINKED ISSUE #489: Number of Products Returned by collection.fetchWithProducts
If you want to get the products that a collection holds you can use the `fetchWithProducts` method:
```javascript
client.collection.fetchWithProducts('Xk9lM2JkNzFmNzIQ4NTIY4ZDFiZTUyZTUwNTE2MDNhZjg==').then((collection) => {
  // Do something with the collection
})
```

This will return onto the `collection` object a 20 products array. There is **no way you can specify the number of products this method should return**.

I propose to add an optional param after `id`, called `pageSize` as the [`product.fetchAll`](http://shopify.github.io/js-buy-sdk/product-resource.js.html#line32) has.

StefanSlehta:
Closes #489

Adds an optional param to filter products on a collection fetched via `fetchWithProducts`. 
Defaults to 20.

### DIFF
--- a/src/collection-resource.js
+++ b/src/collection-resource.js
@@ -78,9 +78,9 @@ class CollectionResource extends Resource {
    * @param {String} id The id of the collection to fetch.
    * @return {Promise|GraphModel} A promise resolving with a `GraphModel` of the collection.
    */
-  fetchWithProducts(id) {
+  fetchWithProducts(id, {productsFirst = 20} = {}) {
     return this.graphQLClient
-      .send(collectionNodeWithProductsQuery, {id})
+      .send(collectionNodeWithProductsQuery, {id, productsFirst})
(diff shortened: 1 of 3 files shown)


src/collection-resource.js
```diff
@@ -78,9 +78,9 @@ class CollectionResource extends Resource {
    * @param {String} id The id of the collection to fetch.
    * @return {Promise|GraphModel} A promise resolving with a `GraphModel` of the collection.
    */
-  fetchWithProducts(id) {
+  fetchWithProducts(id, productsFirst = 20) {
```
minasmart: In the case the signature expands should this not be a key word arg instead of a positional param?

PR #711, opened 2019-10-02
Schema-level errors do not contain data or model properties

sshaw:
This results in a `TypeError`: 

```
TypeError: Cannot read property 'checkoutCreate' of undefined
at index.umd.min.js:25
```

There are similar errors that take place elsewhere (should I open a separate PR or address here?). 

The fixture I've added is a shortened version of the real response:
```json
{
  "errors": [
    {
      "message": "Variable input of type CheckoutCreateInput! was provided invalid value",
      "locations": [
        {
          "line": 1,
          "column": 3030
        }
      ],
      "extensions": {
        "value": {
          "lineItems": {
            "quantity": 1,
            "variantId": "XXlkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8XXDEzNjY3OTg3Ng=="
          }
        },
        "problems": [
          {
            "path": [],
            "explanation": "Invalid global id `XXlkOi8vc2hvcGlmeS9Qcm9kdWN0VmFyaWFudC8XXDEzNjY3OTg3Ng==`"
          }
        ]
      }
    }
  ]
}
```

Let me know if you prefer this.

PR #722, opened 2019-10-31
ensure handleCheckoutMutation prioritizes error handling

craigmichaelmartin:
Without this change, if there is an error (and no data) handleCheckoutMutation throws a TypeError (eg, `Cannot read property 'checkoutCreate' of undefined`) rather than rejecting with the helpful errors information. This is because it tries to handle aspects of the data path (but is unable) before the error path. This change prioritizes the handling of errors.

### DIFF
--- a/src/handle-checkout-mutation.js
+++ b/src/handle-checkout-mutation.js
@@ -1,5 +1,9 @@
 export default function handleCheckoutMutation(mutationRootKey, client) {
   return function({data, errors, model}) {
+    if (!data && errors && errors.length) {
+      return Promise.reject(new Error(JSON.stringify(errors)));
+    }
+
     const rootData = data[mutationRootKey];
     const rootModel = model[mutationRootKey];
 
@@ -11,10 +15,6 @@ export default function handleCheckoutMutation(mutationRootKey, client) {
 
         return rootModel.checkout;
       });
-    }
-
-    if (errors && errors.length) {
-      return Promise.reject(new Error(JSON.stringify(errors)));
     }
 
     if (rootData && rootData.checkoutUserErrors && rootData.checkoutUserErrors.length) {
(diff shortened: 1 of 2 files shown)


src/handle-checkout-mutation.js
```diff
@@ -1,5 +1,9 @@
 export default function handleCheckoutMutation(mutationRootKey, client) {
   return function({data, errors, model}) {
+    if (!data && errors && errors.length) {
```
osamie: Have you thought about scenarios where we return both `data` and `errors`?

craigmichaelmartin: Hi @osamie! Thanks for your review. I have not. I'm not familiar with this code and so was relying on tests to ensure existing cases were covered. I am simply a user the library and saw the TypeError (with the real error being swallowed) and thought I'd make a PR to bring attention to the issue. I am so unfamiliar with the code, that [my first attempt](https://github.com/Shopify/js-buy-sdk/commit/c39aa7fb08037f294663ba6e8b5650bbff6ab5d2) involved the assumption that if there were errors there would be no data, which a test demonstrates is not true.

#711 looks like a reasonable way to solve this (I hadn't seen that PR when I made this one.. bad job out of me!). I would be a bit hesitant about the approach to allow the data path to unfold by defaulting to an empty object. It seems to me that the ideal scenario would be to prioritize catching the invalid case and returning, rather than providing a default just to allow it to journey the data path without erring before getting to the error handling for the invalid case.

The only thing that would still be valuable from this PR is the [change I made to a test](https://github.com/Shopify/js-buy-sdk/pull/722/commits/7ceca71a18419709901d7a2cd9553f6e00269a07#diff-9400fc19b3e72a86fc71e9b744438ee8L521) since this is the test that was should have been failing given the current code, but wasn't. The fix in #711 adds its own test as well, but I think it would make sense to update the existing test so it is accurate to the scenario it is describing ("without data field").

DrTeePot:
This seems stale now, if there's still value in this PR (maybe that additional test you added 😃?) feel free to open a PR off of the current master and we can get that in!

PR #735, opened 2019-12-17
Add compareAtPrice to variant presentment price

stoufa88:
### Description
In a multi currency shop, we need to calculate subtotal discount on the selected currency, by doing `compareAtPrice.amount - price.amount`of a given presentmentPrice.
This will add the compareAtPrice field.


rebeccajfriedman:
Thanks for this contribution! Can you also update the associated fixtures?

stoufa88:
I'm not sure I did it well, please don't hestitate to point other fixtures that should be updated.

PR #752, opened 2020-04-10
Return an empty string if a null or undefined image is passed.

ryanpcmcquen:


ryanpcmcquen:
How do I re-run the CLA check? I signed it.

PR #812, opened 2021-03-11
Update Storefront API to version 2021-01

isamu:
Update SDK version to Storefront API v2021-01 and update schema info.


isamu:
I signed CLA

PR #877, opened 2022-02-02
add clarification of release version in URL

SeeSamRun:
Updating the language to make it more clear that the specific version would be used if a developer wants to rely on an older version. There has been some confusion reported from partners with these example script tags.


andra-la (approved):
💯

PR #903, opened 2022-07-04
Update SDK to use Storefront API v2022-07

melissaluu:
Update SDK to use Storefront API `v2022-07` and alias all instances of `Image.src` to retrieve `Image.url`.


fightingtheboss (approved):
My first time reviewing one of these changes, but given that it's mostly docs changes, I feel reasonably comfortable approving.

Issue #920, opened 2022-10-30
FetchNextPage is not working since v2.17.0


jacksonwangdev:
### Bug details 

**Describe the bug**
I have been using js-buy-sdk library in my React Native application
Before update to v2.17.0, i could fetch all the collection's products by using fetchWithProducts and then FetchNextPage for pagination. But after the update, i get the following:

`[TypeError: Cannot convert undefined value to object]`

**To Reproduce**
Steps to reproduce the behaviour:

The client is intialised as per normal:
```
var Shopify = Client.buildClient({
    domain: Constants.Shopify.url,
    storefrontAccessToken: Constants.Shopify.storefrontAccessToken,
});
```

Call the fetchWithProducts function:
```
fetchProductsByCollectionId: async (id, per_page) => {
        try {
            const responses = await Shopify.collection.fetchWithProducts(id, {productsFirst: per_page});
            return responses.products.map((item) => new Product(item));
        } catch (err) {
            console.log(err);
            return undefined;
        }
   }
```
And then call the FetchNextPage for pagination
```
fetchNextPage: async (list) => { // list is the output of fetchProductsByCollectionId function above
        if (!list) {return undefined;}
        try {
            const responses = await Shopify.fetchNextPage(list);
            if (responses) {
                const responseJson = responses.model;
                return responseJson.map((item) => new Product(item));
            }
            return undefined;
        } catch (err) {
            console.log(err);
            return undefined;
        }
    }
```

**Expected behavior**
fetchNextPage working fine and i can load more data when i'm scrolling down the product list

**Environment (please complete the following information):**
 - OS: MacOS Monteray
 - SDK Version v2.17.0

https://github.com/Shopify/js-buy-sdk/issues/920#tasklist-block-07048f48-c3be-4cd6-a577-03d477369e08

melissaluu:
This is now fixed in the latest SDK `v2.17.1` - thanks for letting us know!

PR #923, opened 2022-11-02
Remove MoneyFragment

### LINKED ISSUE #919: v2.17.0 breaks fetchNextPage?
### Bug details 

**Describe the bug**

I have been using the https://sdks.shopifycdn.com/js-buy-sdk/v2/latest/index.umd.min.js version of the Shopify JS Buy SDK. Before the update to v2.17.0, I could fetch all the products in a store by using FetchAll and then FetchNextPage for pagination. 

After the update, I get the following: 

```
index.umd.min.js:26 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'products')
    at index.umd.min.js:26:142336
    at Array.reduce (<anonymous>)
    at index.umd.min.js:26:142307
```

**To Reproduce**

The client is intialised as per normal (obscuring for privacy):

```
const client = ShopifyBuy.buildClient({
    domain: 'XXXXXX.myshopify.com',
    storefrontAccessToken: 'XXXXXXXXXXXXXXX'
});
```

The code to fetch the products: 

```function getAllProducts(){
    var allProducts = [];
    console.log(("getAllProducts()");
    return new Promise((resolve, reject) =>
    {
        client.product.fetchAll(250).then((products) => {
            products.forEach(el => {
                allProducts.push(el);
            });
    
            client.fetchNextPage(products).then((res) => {
                res.model.forEach(el => {
                    allProducts.push(el);
                });
                console.log("Returning All Products:");
                console.log((allProducts);
                resolve(allProducts);
            });
        });
    })
}
```

As I have said, if I downgrade to version 2.16.1 this command works again, and there is nothing in the change log about fetchNextPage being altered. 

Output from my console - 2.16.1 on the left, 2.17.0 on the right.
fetchAllError

**Expected behavior**

When this code is run, I expect it to get all the products currently in the store (368 at this time) and return them to me in a created array - as it does in version 2.16.1. 

**Environment (please complete the following information):**

-OS: Windows 10
-Browser: Chrome  106.0.5249.119 
-SDK Version: 2.17.0 (error), 2.16.1 (works)
-Languages: HTML5/Javascript

**Additional context**
Thank you for your time, I hope this can be identified and fixed.

**Bug Report Checklist**
- [x] I have read and agree to the [CODE_OF_CONDUCT.md](https://github.com/Shopify/js-buy-sdk/blob/master/CODE_OF_CONDUCT.md)
- [x] I have read the [CONTRIBUTING.md](https://github.com/Shopify/js-buy-sdk/blob/master/CONTRIBUTING.md) guidelines.
- [x] I have provided a detailed description of the bug, including code samples, and/or network data.
- [x] I have provided information about my development environment, including SDK version.

melissaluu:
Currently, `graphql-js-client` will add multiple definitions of the common `MoneyFragment` if the GQL operation contains multiple fragments that uses this common fragment. 

To fix this, we'll remove this common fragment and replace all instances of `MoneyFragment` with the `MoneyV2` object fields.

Fixes #919 and #920

### DIFF
--- a/src/graphql/AppliedGiftCardFragment.graphql
+++ b/src/graphql/AppliedGiftCardFragment.graphql
@@ -1,18 +1,23 @@
 fragment AppliedGiftCardFragment on AppliedGiftCard {
   amountUsed {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   amountUsedV2: amountUsed {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   balance {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   balanceV2: balance {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   presentmentAmountUsed {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   id
   lastCharacters
--- a/src/graphql/CheckoutFragment.graphql
+++ b/src/graphql/CheckoutFragment.graphql
@@ -24,10 +24,12 @@ fragment CheckoutFragment on Checkout {
   requiresShipping
   note
   paymentDue {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   paymentDueV2: paymentDue {
-    ...MoneyFragment
+    amount
+    currencyCode
   }
   webUrl
   orderStatusUrl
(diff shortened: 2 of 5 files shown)


fightingtheboss (approved):
Looks like a straightforward change. I didn't have a chance to 🎩 but it looks like it's a 1:1 replacement for direct access vs using the fragment.

I have to be honest, based on your description I'm not 100% clear of the cause of the error. If these fragments are included in the query, and they contain the "subfragment" for Money, then that subfragment is included multiple times and that causes an exception? The exceptions raised don't point to this being the problem either, which is frustrating.

Should we add a test for the old behaviour and the new one?

PR #987, opened 2024-08-13
Deprecation notices, migration guide and existing queries.

scottdixon:
- Publishes [migration guide](https://github.com/Shopify/js-buy-sdk/tree/sd-deprecation/migration-guide) and [existing queries](https://github.com/Shopify/js-buy-sdk/blob/sd-deprecation/migration-guide/queries).

- Adds deprecation notice to docs:

image

- Adds deprecation notice to readme:

image


johndwells:
Can I ask, is this the direction this library is taking, is it actually being deprecated?

Because we've also found this branch: [sd-cart](https://github.com/Shopify/js-buy-sdk/tree/sd-cart), which migrates to use the new Cart API. We've been testing this branch and have found it works well, though we did submit a [pull request](https://github.com/Shopify/js-buy-sdk/pull/988) last week that adds some missing info about the underlying product of a line item.

So which is it, is this library being abandoned, or can we expect it to continue to be maintained and kept up to date with the latest API changes?

We love this library and really hope it sticks around!

macpham (approved):

thipokch:
I just started using this library and was caught off guard by the lack of a deprecation notice, which I ended up wasting a lot of time. It would also be very helpful to have more context from Shopify on **why this decision was made**. Thank you. 

@johndwells  Do you think it might be worth considering a community-maintained fork?

PR #999, opened 2025-01-06
Checkout API to Cart API migration

juanpprieto:
Public [SDK deprecation notice](https://shopify.dev/changelog/js-buy-sdk-deprecation-notice)

### WHY are these changes introduced?

The main goal of this (final) version of the SDK is to extend the longevity of it's `.checkout` interface by replacing it with an equivalent interface based on the [Cart API](https://shopify.dev/docs/api/storefront/latest/objects/Cart)

### WHAT is this pull request doing?

Replaces the `Checkout API` dependency from all `.checkout` interface methods such as `fetch`, `addLines`, `removeLines` etc. All existing checkout methods now use Cart API equivalents under the hood.

### HOW to test your changes?

Coming soon

#### Checklist

- [x] Added tests
- [ ] Added or updated documentation
- [ ] Validated buy-button-js operation with this SDK


chadwackerman:
This is late and stores are breaking because of the ineptitude and disorganization here. GET IT DONE.

PR #1000, opened 2025-01-16
Checkout API to Cart API migration

kdaviduik:
Public [SDK deprecation notice](https://shopify.dev/changelog/js-buy-sdk-deprecation-notice)

### WHY are these changes introduced?

The main goal of this (final) version of the SDK is to extend the longevity of its `.checkout` interface by replacing it with an equivalent interface based on the [Cart API](https://shopify.dev/docs/api/storefront/latest/objects/Cart)

### WHAT is this pull request doing?

Replaces the `Checkout API` dependency from all `.checkout` interface methods such as `fetch`, `addLines`, `removeLines` etc. All existing checkout methods now use Cart API equivalents under the hood.

#### Checklist

- [x] Added tests
- [x] Added or updated documentation
- [x] Validated buy-button-js operation with this SDK


juanpprieto (approved):
🚀🚀🚀🚀🚀🚀🚀🚀

PR #1009, opened 2025-03-18
Migrate to Github Action release workflow with NPM provenance support

juanpprieto:


package.json
```diff
@@ -4,28 +4,32 @@
   "description": "The JS Buy SDK is a lightweight library that allows you to build ecommerce into any website. It is based on Shopify's API and provides the ability to retrieve products and collections from your shop, add products to a cart, and checkout.",
   "main": "index.js",
   "jsnext:main": "index.es.js",
-  "repository": "git@github.com:Shopify/js-buy-sdk.git",
+  "repository": {
```
kdaviduik, line 7: (threading)

Should we also add the package-lock.json to the .npmignore file? I see that the yarn-lock is npm ignored

package.json
```diff
@@ -4,28 +4,32 @@
...
+    "url": "git+https://github.com/Shopify/js-buy-sdk.git",
+    "type": "git"
+  },
```
kdaviduik, line 10: (threading)

Our `contributing.md` file just states yarn commands, we probably want to update that file too to switch to npm

PR #1011, opened 2025-03-19
Enable changesets and npm github releases

juanpprieto:


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: Run Ce tests
```
kdaviduik:
```suggestion
name: Run CI tests
```

PR #1038, opened 2025-06-17
Update checkout.webUrl to use the same domain as JS Buy SDK v2 

kdaviduik:
Without this change, if the shop's Online Store primary domain isn't configured correctly and pointing to Shopify, some buyers are unable to checkout successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ npm install shopify-buy@3
 **With Yarn:**
 
 ```bash
-yarn install shopify-buy@3
+yarn add shopify-buy@3
 ```
 
 **CDN:**


test/cart-payload-mapper-test.js
```diff
@@ -471,15 +471,45 @@ suite('cart-payload-mapper-test', () => {
...
+      const cart = {
+        checkoutUrl: 'https://shop.com/checkout?_fd=0'
+      };
+
+      const result = mapCartPayload(cart);
+
       assert.strictEqual(result.webUrl, cart.checkoutUrl);
     });
+
+    test('it updates the value of _fd to 0 if it is present but with a different value', () => {
```
juanpprieto, line 504: Looks great, I'd a two more tests!

1. the url has already other params in it (without `_fd`)
2. the url has already other params and `_fd` too

juanpprieto (approved):

PR #1039, opened 2025-06-20
2 discount bug fixes (multiple discounts + shipping discounts)

kdaviduik:
**Nicer diffs if you review commit by commit (and without whitespace diffs!) :)**

1. Fix bug with adding multiple discount codes when not all codes are applicable
    - A discount code can be added to a cart but might not be `applicable`. In that case, the code will appear in the `discountCodes` array, but there won't be any discount allocations.
    - Since we were using discount allocations to determine which discount codes existed, we would end up inadvertently removing discount codes that were added but not `applicable`.
2. Fix bug where a shipping discount could appear as a line item discount (ie: fixes https://github.com/Shopify/js-buy-sdk/issues/1040)
    - Our previous discount mapping logic did not properly map shipping discounts - they were treated the same as order-level discounts, but behave differently. 
    - Unlike order-level discounts (which must get mapped onto line items), shipping discounts should ONLY appear in the `checkout.discountApplications` array.
    - Importantly, if a shipping discount has been added but there is no shipping address known, the Cart APIs behave differently than the Checkout APIs and we therefore have no way of knowing the value, target type, etc. This information was returned by the Checkout APIs in the `checkout.discountApplications` array, but now we cannot return shipping discounts in that array if we don't have a shipping address.

### DIFF
--- a/src/checkout-resource.js
+++ b/src/checkout-resource.js
@@ -201,20 +201,41 @@ class CheckoutResource extends Resource {
    * @return {Promise|GraphModel} A promise resolving with the updated checkout.
    */
   addDiscount(checkoutId, discountCode) {
-    return this.fetch(checkoutId).then((checkout) => {
-      const existingRootCodes = checkout.discountApplications.map(
-        (discountApplication) => discountApplication.code
-      );
+    // We want access to Cart's `discountCodes` field, so we can't just use the
+    // existing `fetch` method since that also maps and removes the `discountCodes` field.
+    // We must therefore look at the raw Cart data to be able to see ALL existing discount codes,
+    // whether they are `applied` or not.
 
-      const existingLineCodes = checkout.lineItems.map((lineItem) => {
-        return lineItem.discountAllocations.map(
-          ({discountApplication}) => discountApplication.code
-        );
+    // The query below is identical to the `fetch` method's query EXCEPT we don't call `mapCartPayload` here
+    return this.graphQLClient.send(cartNodeQuery, {id: checkoutId}).then(({model, data}) => {
+      return new Promise((resolve, reject) => {
+        try {
+          const cart = data.cart || data.node;
+
+          if (!cart) {
+            return resolve(null);
+          }
+
+          return this.graphQLClient
+            .fetchAllPages(model.cart.lines, {pageSize: 250})
+            .then((lines) => {
+              model.cart.attrs.lines = lines;
+
+              return resolve(model.cart);
+            });
+        } catch (error) {
+          if (error) {
+            reject(error);
+          } else {
+            reject([{message: 'an unknown error has occurred.'}]);
+          }
+        }
+
+        return resolve(null);
       });
-
-      // get unique applied codes
-      const existingCodes = Array.from(
-        new Set([...existingRootCodes, ...existingLineCodes.flat()])
+    }).then((checkout) => {
+      const existingCodes = checkout.discountCodes.map(
+        (code) => code.code
       );
 
       const variables = this.inputMapper.addDiscount(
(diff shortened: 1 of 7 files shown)


src/utilities/cart-discount-mapping.js
```diff
@@ -152,44 +152,84 @@ export function discountMapper({cartLineItems, cartDiscountAllocations, cartDisc
...
-    // Check if the code exists in the map (case-insensitive)
-    let found = false;
-
-    for (const [key] of discountIdToDiscountApplicationMap) {
-      if (key.toLowerCase() === code.toLowerCase()) {
-        found = true;
-        break;
-      }
-    }
-    if (!found) {
```
kdaviduik, line 183: Deleting this because this scenario is now **expected** if we have a shipping discount code added but no shipping address information

src/utilities/cart-discount-mapping.js
```diff
@@ -231,86 +231,90 @@ function generateDiscountApplications(cartLinesWithAllDiscountAllocations, disco
     if (!discountAllocations) { return; }
 
     discountAllocations.forEach((discountAllocation) => {
-      const discountApp = discountAllocation.discountApplication;
-      const discountId = getDiscountAllocationId(discountAllocation);
+      createCheckoutDiscountApplicationFromCartDiscountAllocation(discountAllocation, discountIdToDiscountApplicationMap, discountCodes);
```
kdaviduik, line 274: Review without whitespace diffs! 
image

test/client-checkout-discounts-integration-test.js
```diff
@@ -1189,7 +1190,7 @@ suite('client-checkout-discounts-integration-test', () => {
                 );
               });
           });
-      });
+      }).timeout(5000);
```
kdaviduik, line 1193: This was failing sometimes due to timeouts

test/client-checkout-discounts-integration-test.js
```diff
@@ -1763,82 +1764,6 @@ suite('client-checkout-discounts-integration-test', () => {
           });
       });
 
-      test(
```
kdaviduik, line 1766: added equivalent test below

test/client-checkout-discounts-integration-test.js
```diff
@@ -1981,5 +1906,525 @@ suite('client-checkout-discounts-integration-test', () => {
...
+        expectedRootDiscountApplications.sort((a, b) => a.code.localeCompare(b.code));
+        updatedCheckout.discountApplications.forEach((discountApplication, index) => {
+          assertActualDiscountApplicationIsExpected(
+            discountApplication,
+            expectedRootDiscountApplications[index]
+          );
+        });
+      }).timeout(5000);
+
+      test('gracefully returns an empty array for discountApplications when adding a free shipping discount to a checkout with a shipping address but no line items', async() => {
```
kdaviduik, line 2110: This one was a bit unexpected but turns out this is the actual Checkout API behaviour

src/checkout-resource.js
```diff
@@ -201,20 +201,41 @@ class CheckoutResource extends Resource {
...
+          }
+        }
 
-      // get unique applied codes
-      const existingCodes = Array.from(
-        new Set([...existingRootCodes, ...existingLineCodes.flat()])
+        return resolve(null);
+      });
+    }).then((checkout) => {
+      const existingCodes = checkout.discountCodes.map(
```
juanpprieto, line 237: How come we no long have to filter out the uniquely applied codes as we did before?

test/client-checkout-discounts-integration-test.js
juanpprieto: 😑 hah

src/checkout-resource.js
kdaviduik: The `discountCodes` array returned from the cart is the **set** of discount codes applied, so we don't have to do anything.

Whereas before we were extracting the unique discount codes from the discount allocations, and there can be multiple discount allocations for a single discount code

PR #1044, opened 2025-07-10
[DOCS] Delete .github/dependabot.yml

rxbchen:
Delete the extra dependabot.yml that was added by accident


kdaviduik (approved):
Updated PR title so that it won't trigger an npm release

